PR #173, opened 2018-04-10
gRPC Code Sync

stanley-cheung:


net/grpc/gateway/runtime/nginx_notify_queue.cc
```diff
@@ -19,6 +19,7 @@
 #include "net/grpc/gateway/runtime/nginx_notify_queue.h"
 
 #include <algorithm>
+#include <utility>
```
fengli79: Why add this?

net/grpc/gateway/runtime/tag.cc
```diff
@@ -19,6 +19,7 @@
 #include "net/grpc/gateway/runtime/tag.h"
 
 #include <algorithm>
+#include <utility>
```
fengli79: Same question.

PR #185, opened 2018-05-14
Support for proto files without packages

### LINKED ISSUE #184: [protoc-gen-web-grpc] Invalid output from proto files without package specified
A proto file that does not specify a package would result in invalid js files that contain similar to this:

```JavaScript
proto..TestServiceClient = /* ... */
```

zaucy:
This fixes #184


stanley-cheung:
ok to test

PR #193, opened 2018-06-07
Add Envoy config file

stanley-cheung:
Main changes this PR:
 - Add an Envoy config file envoy.yaml, which will allow us to run the end-to-end Echo example swapping out Nginx with Envoy.
 - Replace references to Nginx with Envoy in documentations as much as possible
 - Split the end-to-end Echo examples into 4 Dockerfiles to make it more modular
    - prereqs, echo_server, envoy and static_assets
 - Add a docker-compose.yml file to further simplify the end-to-end example

To run:
- docker-compose up
- hit http://localhost/net/grpc/gateway/examples/echo/echotest.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,20 +4,20 @@ gRPC-Web provides a Javascript client library that lets browser clients
 access a gRPC server. You can find out much more about gRPC in its own
 [website](https://grpc.io).
 
-The current release is a Beta release, and we expect to announce General-Available by Oct. 2018. 
+The current release is a Beta release, and we expect to announce
+General-Available by Oct. 2018. 
 
-The JS client library has been
-used for some time by Google and Alphabet projects with the
+The JS client library has been used for some time by Google and Alphabet
+projects with the
 [Closure compiler](https://github.com/google/closure-compiler)
 and its TypeScript generator (which has not yet been open-sourced).
 
 gRPC-Web clients connect to gRPC servers via a special gateway proxy: our
-provided version uses Nginx. 
+provided version uses [Envoy](https://github.com/envoyproxy/envoy), in which
+gRPC-Web support is built-in. Envoy will become the default gateway for
+gRPC-Web by GA.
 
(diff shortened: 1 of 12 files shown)


glerchundi:
@stanley-cheung I don't know if this the correct place to make a note but I would like to say that gRPC-Web is already supported by default with Envoy + Heptio Countour: https://github.com/heptio/contour/pull/319.

Of course this is for people who are using Kubernetes...if not, please start using it :)

README.md
```diff
@@ -4,20 +4,19 @@ gRPC-Web provides a Javascript client library that lets browser clients
...
-The JS client library has been
-used for some time by Google and Alphabet projects with the
+The JS client library has been used for some time by Google and Alphabet
+projects with the
 [Closure compiler](https://github.com/google/closure-compiler)
 and its TypeScript generator (which has not yet been open-sourced).
 
 gRPC-Web clients connect to gRPC servers via a special gateway proxy: our
-provided version uses Nginx. 
+provided version uses [Envoy](https://github.com/lyft/envoy), in which gRPC-Web
```
ctaggart: That redirects to `https://github.com/envoyproxy/envoy` now

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      dockerfile: ./net/grpc/gateway/docker/echo_server/Dockerfile
+    image: grpc-web:echo-server
+    ports:
+      - "9090:9090"
+  envoy:
+    build:
+      context: ./
+      dockerfile: ./net/grpc/gateway/docker/envoy/Dockerfile
+    image: grpc-web:envoy
+    network_mode: "host"
```
neilcarpenter: Using `network="host"` is not supported on docker for Mac (see https://docs.docker.com/network/host/).

Using [`link`](https://docs.docker.com/compose/compose-file/#links) to connect services to envoy and then referencing service name in the envoy config works fine though.

PR #199, opened 2018-06-19
Include advantages of solution in README

benmccann:


README.md
```diff
@@ -21,6 +21,11 @@ In future, we expect gRPC-Web to be supported in
 language-specific Web frameworks, such as Python, Java, and Node. See the
 [roadmap](https://github.com/grpc/grpc-web/blob/master/ROADMAP.md) doc.
 
+Some of the advantages of Protocol Buffers over JSON are:
+
+* Less data to communicate same information
+* Native classes generated for both client and server (so structure and types are built-in)
```
thomasetter: I would also mention the possibility to rename fields and change compatible types as that's not possible with the common JSON approach of storing field names in the key.
It would obviously also be possible to do renames by using the tag number as a key, but that's not what the standard [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) does.

wenbozhu:
Thanks for all the inputs! 

I am not sure we want to talk about protobuf over JSON on the main doc of grpc-web. Could use a background section, and then we should also mention disadvantages? ;)

Specifically to what you have mentioned:
1) JSON is larger on wire (and compression may change the delta) but a lot faster in parsing speed. Code-size is also an issue for protobufs. We have a solution to get the best of both worlds, but open-sourcing it requires quite a bit of work. Binary data is also an issue for security, so base64 encoding is sometimes necessary which also adds overhead.
2) Typed APIs, yes, definitely the biggest value provided by protobuf.
3) Wire compatibility, yes, this is the biggest reason not to use JSON (unless you version your APIs)

If you feel strongly about this, I am happy to draft or review a dedicate doc on this subject. It will offer some very useful background for grpc-web users.

benmccann:
Thanks. I agree with mentioning the tradeoffs in both directions. Engineering is all about understanding what tradeoffs you're making. I think we could be much clearer about why someone would include grpc-web in their stack and help them make that decision. It would help adoption of this project a lot to mention some of the benefits

I would happily review a doc if you were able to draft one. Unfortunately I don't think I have the background to be able to draft myself. I was only including information I got from various maintainers of the project. I'm not sure we need an entirely separate doc and thought we could just place a few bullets such as the ones below. But if you wanted to create a document I certainly wouldn't complain :-)

Protocol Buffers compared to JSON:

* Use less data to communicate same information
* Generate native classes for both client and server (so structure and types are built-in)
* Offer the ability to rename fields and change compatible types
* Are slower to parse

johanbrandhorst:
Just jumping in here, I assume "protobuf is slower to parse" is only true in the context of the browser, and even then I'd love to see some hard numbers on that assertion.

wenbozhu:
@johanbrandhorst  of course, the context for this whole discussion is for web (HTML5) clients.

Speed is only an issue for web apps that need process a lot of data, e.g. think ML.

Code size is only an issue for web apps that share a lot of state between client and server, e.g. "gmail", which is usually not a case for API-based apps.

===

gRPC is agnostic to serialization-formats .. i.e. why I like to avoid discussing such a broad topic on the intro page of grpc-web.

benmccann:
@wenbozhu gRPC is agnostic to serialization-formats, but is grpc-web? I thought grpc-web only supports protobuf, but I'm not very familiar with this project, so could have gotten something wrong

johanbrandhorst:
@benmccann the existing spec mentions both JSON and protobuf as transport formats, so yes, gRPC-Web is agnostic.

johanbrandhorst:
See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2

benmccann:
What would the motivation be for using grpc-web with json? It doesn't seem like you'd need a client library if the responses are coming back in json because the browser already has native support. Or does grpc-web still generate a typed object for you even in that case?

johanbrandhorst:
Of course. The JSON would just be on the transport level, you'd still have generated types marshalled to and from the JSON representation. Some web developers prefer seeing exactly the contents of their messages "on the wire" (the browser network tab).

benmccann:
Maybe we could say something like the following on the README:

gRPC-web generates ECMAScript objects and provides marshaling to and/from those types from several different transport protocols including protobuf, thrift, and json.

wenbozhu:
Proto3 supports canonical JSON mapping (i.e. JSON being a supported wire format). However, the JS protobuf library grpc/grpc-web uses doesn't yet support JSON (as we used a more compact JSON format internally for web apps).

===

gRPC as an RPC framework may support other serialization formats such as Thrift, but this is less relevant to grpc-web. 

I will propose some clarification on the readme.md.  A full discussion on protobuf v.s. JSON could come later.

Issue #197, opened 2018-06-16
feature request: grpc-web client in commonjs + npm package


glerchundi:
We found lots of issues when trying to integrate the current gRPC Web Client implementation into our stack because we (and probably most of the non-Googlers) don't use Google Closure for nothing.

I would like to see the client library released in npm and commonjs compliant which probably would solve all the issues we've had. Javascript Protocol Buffers generator already supports commonjs so it seems like a good idea to have a re-implementation of the client in this way.

WDYT? Is this on the roadmap?

jonahbron:
My project is using Improbable's GRPC client and proxy (CommonJS), and it's working really well.  Might pay off to use it until this official implementation matures a bit more for external use.

glerchundi:
@jonahbron thanks for the pointers but we prefer to stick with an official development of grpc-web although it's in closure instead of relying on a third-party implementation.

n0mer:
@wenbozhu good news, thanks - looking forward as well!

stanley-cheung:
May I ask the community what the expectation for this feature request is?

1. Right now, the JS code in the core / generic gRPC-Web library (i.e. stuff in this directory https://github.com/grpc/grpc-web/tree/master/javascript/net/grpc/web) uses the `goog.module`, `good.require` include style. I can certainly make them commonjs style (i.e. `require`, `module.exports`). Is that what's expected here?

2. How does the protoc plugin generated code fit in? The current expectation is that, you have a `.proto` definition.   2a.  You run our [gRPC-Web protoc plugin](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpc_generator.cc). That generates the gRPC-Web API .JS client stub.  2b.   With this, and code above, you run the [`closuer-compiler.jar`](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/Makefile#L65-L73) to generate one single, compiled, JS file which has no other runtime dependencies. 

With the output of 2b, you are free to embed that into whatever frontend framework you will be using. Step 2a and 2b should be mostly 1-time build step. How does the gRPC-Web core library code being a npm module help in this case?

I may have missed something obvious so please let us know how you guys think this should work and I will be happy to help.

zaucy:
Having the option for the import style to be commonjs and having all common-functionality in an npm module is what my expectation would be. Here's a simple example:

```JavaScript
import {StatusCode} from 'grpc-web';

// ...

[...]

PR #210, opened 2018-07-27
grpcweb npm runtime module

zaucy:
Created for #197 

This sets up an npm module that runs the google closure compiler on the client related js before publishing to npm.

### DIFF
--- a/packages/grpc-web/.gitignore
+++ b/packages/grpc-web/.gitignore
@@ -1,0 +1,3 @@
+/index.js
+/package-lock.json
+/node_modules
--- a/packages/grpc-web/.npmignore
+++ b/packages/grpc-web/.npmignore
@@ -1,0 +1,3 @@
+/scripts
+/package-lock.json
+/.gitignore
--- a/packages/grpc-web/README.md
+++ b/packages/grpc-web/README.md
@@ -1,0 +1,58 @@
+# gRPC-Web
+
+gRPC-Web provides a Javascript client library that lets browser clients
+access a gRPC server. You can find out much more about gRPC in its own
+[website](https://grpc.io).
+
+The current release is a Beta release, and we expect to announce
+General-Available by Oct. 2018.
+
+The JS client library has been used for some time by Google and Alphabet
+projects with the
+[Closure compiler](https://github.com/google/closure-compiler)
+and its TypeScript generator (which has not yet been open-sourced).
+
+gRPC-Web clients connect to gRPC servers via a special gateway proxy: our
+provided version uses [Envoy](https://github.com/envoyproxy/envoy), in which
+gRPC-Web support is built-in. Envoy will become the default gateway for
+gRPC-Web by GA.
+
+In the future, we expect gRPC-Web to be supported in
+language-specific Web frameworks, such as Python, Java, and Node. See the
+[roadmap](https://github.com/grpc/grpc-web/blob/master/ROADMAP.md) doc.
+
+## Quick start
+
+This example is using the `echo.proto` file from the [Echo Example](net/grpc/gateway/examples/echo).
+
+1. Add `grpc-web` as a dependency using `npm` or `yarn`.
+
+    ```shell
+    npm i grpc-web
+    ```
+
+    ```shell
+    yarn add grpc-web
+    ```
+
+2. Generate your client with `protoc` and the `protoc-gen-grpc-web` plugin. Make sure you set the import_style for both `js_out` and `grpc-web_out` to **commonjs**. It is also important that both your js and grpc-web output to the same directory.
+
+    ```shell
+    protoc echo.proto --js_out=import_style=commonjs:generated --grpc-web_out=import_style=commonjs,mode=grpcweb,out=echo_grpc_pb.js:generated
+    ```
+
+3. Start using your generated client!
+
+    ```js
+    const {EchoRequest} = require('./generated/echo_pb.js');
+    const {EchoServiceClient} = require('./generated/echo_grpc_pb.js');
+
+    const client = new EchoServiceClient('localhost:9090');
+
+    const request = new EchoRequest();
+    request.setMessage('Hello World!');
+
+    client.echo(request, {}, (err, response) => {
+      // ...
+    });
+    ```
--- a/packages/grpc-web/package.json
+++ b/packages/grpc-web/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "grpc-web",
+  "version": "0.1.0",
+  "description": "gRPC web runtime",
+  "main": "index.js",
+  "scripts": {
+    "prepublishOnly": "npm run build",
+    "build": "node scripts/build.js"
+  },
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "google-closure-compiler": "^20180716.0.0"
+  }
+}
--- a/packages/grpc-web/scripts/build.js
+++ b/packages/grpc-web/scripts/build.js
@@ -1,0 +1,42 @@
+const path = require("path");
+const {exec} = require("child_process");
+
+const cwd = process.cwd();
+
+const indexPath = path.relative(cwd, path.resolve(__dirname, "../index.js"));
+
+const jsPaths = [
+  "../../../javascript",
+  "../../../third_party/closure-library",
+  "../../../third_party/grpc/third_party/protobuf/js",
+].map(jsPath => path.relative(cwd, path.resolve(__dirname, jsPath)));
+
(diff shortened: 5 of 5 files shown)


npm/scripts/build.js
```diff
@@ -0,0 +1,46 @@
...
+  "grpc.web.ClientReadableStream",
+  "grpc.web.Error",
+  "grpc.web.GrpcWebClientReadableStream",
+  "grpc.web.GrpcWebStreamParser",
+  "grpc.web.Status",
+  "grpc.web.StatusCode",
+  "grpc.web.StreamBodyClientReadableStream",
+  "grpc.web.GrpcWebClientBase",
+
+  // @TODO: Include these
```
stanley-cheung: Doesn't really need to. These 2 files are irrelevant for this purpose.

zaucy: Wouldn't `grpc.web.GatewayClientBase` be required if a client was compiled with `mode=base64`?

stanley-cheung: We may have to remove it for some other reasons in the near future. So I just don't want to call this out as a `TODO`. We can add support for that any time.

stanley-cheung:
Is the `package-lock.json` necessary to be checked in? It seems to be huge. Is it all just pulled in by the `google-closure-compiler` dependency (and then recursive for other `node_modules`)?

stanley-cheung:
Should `package.json` be on the root directory of the repo instead?

`.npmignore` can be put in the root directory as well.

`.gitignore` can merge with the one in root directory (seems to be introduced by 3 PRs at the same time :P) 

`script/build.js` can be put into the existing `scripts` directory

I am not sure if the `npm/` folder is a convention?

zaucy:
Typically it would be included in a project to keep its runtime dependencies from changing between installs, but in this case it actually is unncessary. There are no runtime dependencies, just the `google-closure-compiler` dev dependency. It was really just force of habit. I'll remove it.

stanley-cheung:
Btw, `index.js` has not been checked in?

zaucy:
> Btw, `index.js` has not been checked in?

It gets generated on an `npm publish`

stanley-cheung:
Ah. Can you add some notes to the main `README.md` to show how a typical npm workflow goes if a user wants to use grpc-web thru that route? Perhaps a `Quickstart with npm` section or something like that.

Thanks for your contribution! This is very important.

zaucy:
It would mean that all the other files/folders will need to be added to the `.npmignore` or else they would be included with a publish. Is that okay?

lalomartins:
Please make sure to include the Readme (or *a* Readme) in the package so that it gets displayed on npmjs.com; looking at your test package at the moment gives me “Unable to find a readme“ which is not ideal.

stanley-cheung:
> It would mean that all the other files/folders will need to be added to the .npmignore or else they would be included with a publish. Is that okay?

Yes please add `.npmignore` to all the other folders. Thanks.

npm/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "grpc-web",
+  "version": "0.0.0",
```
stanley-cheung: Let's call it version 0.1.0 please. Thanks.

stanley-cheung:
> > It would mean that all the other files/folders will need to be added to the .npmignore or else they would be included with a publish. Is that okay?

> Yes please add .npmignore to all the other folders. Thanks.

Now that I think about it more. Should we put `package.json` and `.npmignore` only at the `javascript/net/grpc/web` directory? 

This repo contains both the gRPC-Web client JS code and the custom Nginx proxy code. `package.json` obviously doesn't apply to the latter. So it looks like `package.json` belong better at `javascript/net/grpc/web/` folder?

Sorry for going back and forth. Not very familiar with how github repo and npm modules interact

lalomartins:
@stanley-cheung Normally they correspond one-to-one. But in cases where they don't, especially the “monorepo” pattern, indeed `package.json` and friends go in the inner directory (in the monorepo case, `packages/foo`). So that would be a +1 from me, but it would still be preferable to somehow include the Readme, and that would require some extra gymnastics in this case.

zaucy:
Is `javascript/net/grpc/web/` the final decision?

I'm more in favour of `packages/grpc-web` myself. I could see another npm package in this repository that wraps the protoc plugin in an npm module which could reside in `packages/protoc-gen-grpc-web`.

It may make sense to maintain separate README's for the npm module instead of using the main repository one. The quick start using docker-compose for instance could be omitted as well as any closure related examples. A project that is using the npm module most likely doesn't need to see that information first hand while viewing the package at npmjs.com, yarnpkg.com etc.

stanley-cheung:
+1 to `packages/grpc-web`. This is how `grpc/grpc-node` repo does it - they have multiple npm packages under the `packages/` folder.

lalomartins:
Well I'm not a contributor, I'm not CNCF and not Google, so anything I say is from the perspective of an user with long years of JS experience.

From that perspective, I'd also prefer packages/grpc-web. It's more intuitive and more in line with the community standards. **But**, in that case the correct thing to do would be to move the actual sources to that location, and that would mean changing the whole build process for the Closure version as well. I'm not sure how the maintainers would like that.

stanley-cheung:
I am open to moving all the JS source code to be under `packages/grpc-web`. The code being in `javascript/net/grpc/web` was purely because we started developing code for internal use case first. This should have no bearing for the OSS/Github use case. 

@zaucy If you feel like moving all the source code is too big a task, feel free to leave the PR in the current state and I can do the big move on top of that.

stanley-cheung:
Actually we just need to rename `javascript/net/grpc/web` to `packages/grpc-web`?

zaucy:
I kept the scripts folder for the `grpc-web` module in `packages/grpc-web/scripts` just because the script is only related to the build step when running `npm run build`.

> Actually we just need to rename javascript/net/grpc/web to packages/grpc-web?

I'll let you move the js files in case there are any issues with the examples / closure related pieces that I haven't touched on. All that would need to be changed in the grpc-web package is this line https://github.com/grpc/grpc-web/pull/210/files#diff-dabf4379a467e2218cfdb99795a1558bR9

zaucy:
Is `packages/grpc-web/README.md` okay? I just took the top part of the main README and added a quick start using protoc.

stanley-cheung:
> Is packages/grpc-web/README.md okay? I just took the top part of the main README and added a quick start using protoc.

Yep SGTM.

@zaucy Thanks for all the great work! Now all the changes are in the new `packages/grpc-web` folder. Very clean. I will move the rest of the source code later (and make sure all the other dockerfiles and stuff are matched).

stanley-cheung:
@zaucy Can you squash the commits please?

zaucy:
@stanley-cheung Done!

stanley-cheung:
I got an error when running `npm run build` from the `packages/grpc-web` directory just now

```
~/grpc-web/packages/grpc-web$ npm run build

> grpc-web@0.1.0 build /home/stanleycheung/grpc-web/packages/grpc-web
> node scripts/build.js

/home/stanleycheung/grpc-web/packages/grpc-web/scripts/build.js:2
const {exec} = require("child_process");
      ^

SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3

npm ERR! Linux 4.13.0-1019-gcp
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "run" "build"
npm ERR! node v4.2.6
npm ERR! npm  v3.5.2
npm ERR! code ELIFECYCLE
npm ERR! grpc-web@0.1.0 build: `node scripts/build.js`
npm ERR! Exit status 1
```

zaucy:
Ah the script is not compatible with your version of node. Is it required that it work with node 4? I'm running node 8 which is the current LTS.

lalomartins:
Node 4 is way past EOL. The earliest supported release is 6.

I guess this is an use case for the `engines` field in `package.json`?

zaucy:
> I guess this is an use case for the engines field in package.json?

That would be unnecessary. The error is caused by the build script. The `engines` field would be used for end users installing the package.

stanley-cheung:
Great! This all works. I ran `npm run build` from the `packages/grpc-web` directory and I see that the `index.js` file got compiled by the closure compiler.

PR #211, opened 2018-07-27
commonjs support with import_style option

zaucy:
Created for #197 

Works in tandem with #210

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -47,6 +47,11 @@ enum Mode {
   GRPCWEB = 4,     // client using the application/grpc-web wire format
 };
 
+enum ImportStyle {
+  CLOSURE = 0,  // goog.require("grpc.web.*")
+  COMMONJS = 1, // const grpcWeb = require("grpc-web")
+};
+
 string GetModeVar(const Mode mode) {
   switch (mode) {
     case OP:
@@ -118,6 +123,34 @@ void PrintMessagesDeps(Printer* printer, const FileDescriptor* file) {
         "goog.require('proto.$full_name$');\n");
   }
   printer->Print("\n\n\n");
+}
+
+void PrintCommonJsMessagesDeps(Printer* printer, const FileDescriptor* file) {
+  std::map<string, const Descriptor*> messages = GetAllMessages(file);
+  std::map<string, string> vars;
+  string package = file->package();
+  string filename = file->name();
+  // Remove .proto extension
+  filename = filename.substr(0, filename.size() - 6);
+  vars["package_name"] = package;
+  vars["filename"] = filename;
+
+  printer->Print(vars, "const proto = {};\n");
+  if (!package.empty()) {
+
+    size_t offset = 0;
+    size_t dotIndex = package.find('.');
+
+    while (dotIndex != string::npos) {
+      vars["current_package_ns"] = package.substr(0, dotIndex);
+      printer->Print(vars, "proto.$current_package_ns$ = {};\n");
+
+      offset = dotIndex + 1;
+      dotIndex = package.find(".", offset);
+    }
+  }
+
+  printer->Print(vars, "proto.$package_name$ = require('./$filename$_pb.js');\n\n");
 }
 
 void PrintFileHeader(Printer* printer, const std::map<string, string>& vars) {
@@ -290,16 +323,22 @@ class GrpcCodeGenerator : public CodeGenerator {
 
     string file_name;
     string mode;
-    for (uint i = 0; i < options.size(); ++i) {
+    string import_style_str;
+    ImportStyle import_style;
+  
+    for (size_t i = 0; i < options.size(); ++i) {
       if (options[i].first == "out") {
         file_name = options[i].second;
       } else if (options[i].first == "mode") {
         mode = options[i].second;
+      } else if (options[i].first == "import_style") {
+        import_style_str = options[i].second;
       } else {
         *error = "unsupported options: " + options[i].first;
         return false;
       }
     }
+
     if (file_name.empty()) {
       *error = "options: out is required";
       return false;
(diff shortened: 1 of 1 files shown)


javascript/net/grpc/web/grpc_generator.cc
```diff
@@ -119,6 +124,34 @@ void PrintMessagesDeps(Printer* printer, const FileDescriptor* file) {
...
+  std::map<string, string> vars;
+  string package = file->package();
+  string filename = file->name();
+  // Remove .proto extension
+  filename = filename.substr(0, filename.size() - 6);
+  vars["package_name"] = package;
+  vars["filename"] = filename;
+
+  printer->Print(vars, "const proto = {};\n");
+  if(!package.empty()) {
```
stanley-cheung: Can you add a `<space>` after each `if`, `while`, and `switch` for consistency please?

stanley-cheung:
You might have to rebase this as well. Thanks

javascript/net/grpc/web/grpc_generator.cc
```diff
@@ -290,16 +323,22 @@ class GrpcCodeGenerator : public CodeGenerator {
 
     string file_name;
     string mode;
+    string import_style_str;
+    ImportStyle import_style;
+  
     for (uint i = 0; i < options.size(); ++i) {
```
zaucy: @stanley-cheung This is causing an error for me. Should this not be `size_t`?

stanley-cheung: Ah right. Sorry please change.

zaucy: Done!

javascript/net/grpc/web/grpc_generator.cc
```diff
@@ -120,6 +125,34 @@ void PrintMessagesDeps(Printer* printer, const FileDescriptor* file) {
...
+    while (dotIndex != string::npos) {
+      vars["current_package_ns"] = package.substr(0, dotIndex);
+      printer->Print(vars, "proto.$current_package_ns$ = {};\n");
+
+      offset = dotIndex + 1;
+      dotIndex = package.find(".", offset);
+    }
+  }
+
+  printer->Print(vars, "proto.$package_name$ = require('./$filename$_pb.js');\n\n");
```
stanley-cheung: Does this mean that the proto messages classes have to be generated into one single file named <some name>_pb.js, and can't be generated into one file per message?

zaucy: Yeah I'm matching how the protoc compiler outputs it's message with import_style set to commonjs. They output a js file per proto file with the name `<proto-file-name>_pb.js`.

stanley-cheung:
@zaucy Do you mind doing a rebase here?

I am still trying to run the full npm + commonjs example. Just to make sure: you are using the same Echo backend gRPC c++ server + Envoy proxy right? So the only thing I need to swap out is how the frontend HTML calls the JS code right? I suppose one easy way is just set my webserver to point to a directory with index.js + node_modules/ generated from PR #210?

zaucy:
@stanley-cheung Done!

Yeah all you should have to do is change how the frontend imports the client code. You'll need a frontend that works with commonjs imports. Anything based off webpack would work. Do you want me to write an example in webpack, angular and react? I can do that in another PR early next week.

stanley-cheung:
@zaucy Yea such a complete example would greatly benefit the project! Thanks again for your great contributions!

stanley-cheung:
Ha, I finally was able to (well, use a very roundabout way) finally verify that the CommonJS generated code works end-to-end!

Hack 1: I put all the generated code `echo_pb.js` and `echo_grpc_pb.js` into the same directory where `package.json` is, in the `packages/grpc-web` folder. And manually hack all the `require` to be like `require('./echo_pb.js');`

Hack 2: This line in the grpc-web generated code doesn't work
`grpc.web = require('grpc-web');`

But because I ran `npm run build` before, I changed that line to
`grpc.web = require('./index.js');`

And then I make `client.js`
```
const {EchoRequest} = require('./echo_pb.js');                                                                
const {EchoServiceClient} = require('./echo_grpc_pb.js');                                                     
                                                                                                              
const client = new EchoServiceClient('http://localhost:8080', null, null);                                    
                                                                                                              
const request = new EchoRequest();                                                                            
request.setMessage('Hello World!');                                                                           
                                                                                                              
client.echo(request, {}, (err, response) => {                                                                 
  console.log(response.getMessage());                                                                         
});  
```

And then `./node_modules/.bin/browserify client.js > out.js`

And finally an `index.html` file that has a `<script>` tag that runs `out.js`


Is it because I haven't "published" the `grpc-web` module to npm that I need to hack `require('grpc-web');` to `require('./index.js');`?

stanley-cheung:
@zaucy I noticed a potential discrepancy on where the generated file is placed for the grpc plugin vs the normal (message) protoc JS plugin:

Let's say my protoc command is like this:
```
~/grpc-web$ protoc ./net/grpc/gateway/examples/echo/echo.proto --js_out=import_style=commonjs:generated --plugin=protoc-gen-grpc-web=./javascript/net/grpc/web/protoc-gen-grpc-web --grpc-web_out=import_style=commonjs,mode=grpcweb,out=echo_grpc_pb.js:generated
```

The grpc generated file is placed in `./generated/echo_grpc_pb.js`, but the messages (generated by the standard protoc JS plugin) file is placed in `./net/grpc/gateway/examples/echo/echo_pb.js`.

Should they be both placed in the `./net/grpc/gateway/examples/echo` directory?

lalomartins:
Off the top of my head generators for other languages (I've worked with Python and C# at least, maybe C++) put all output relative to where the command is run, it this case `./generated`. Given the way I'd expect to automate generating bindings for multiple languages, my gut feeling is that's the right behavior, and the `pb.js` file is in the wrong.

And the reasoning is simple: if you want it in the same place as the `proto` file, that's easy to achieve by running the command with that cwd. In fact, the command line explicitly asks for `:generated` as the output path, but that's ignored, so I'd call it a bug.

zaucy:
@stanley-cheung  I'll take a look later today and make sure it's consistent.

stanley-cheung:
@lalomartins sorry I was a little ambiguous. The `_pb.js` still honors the `./generated`. I edited my post.

One is `./generated/echo_grpc_pb.js`. The other is `./generated/net/grpc/gateway/examples/echo/echo_pb.js`

lalomartins:
Ah, I see what you mean. That's especially important if there are multiple files and they're not all in the same place (like you have a subdirectory — we had a layout like this in [a recent project](https://github.com/listenfield/listenfield-apis/tree/master/protos)).

stanley-cheung:
Nevermind, I can manipulate the `protoc` command a bit and get the generated files into the same directory.

```
protoc -I=net/grpc/gateway/examples/echo echo.proto
--plugin=protoc-gen-grpc-web=./javascript/net/grpc/web/protoc-gen-grpc-web
--js_out=import_style=commonjs:generated
--grpc-web_out=import_style=commonjs,mode=grpcweb,out=echo_grpc_pb.js:generated
```

PR #219, opened 2018-08-04
Add bazel rule closure_grpc_web_library

Yannic:


stanley-cheung:
Wow, this is great! Thanks for the contribution! I will take a closer look shortly. Will let test run first.

bazel/closure_grpc_web_library.bzl
```diff
@@ -0,0 +1,195 @@
...
+      transitive_sources = dep.proto.transitive_imports,
+  )
+
+  deps = unfurl(ctx.attr.deps, provider = "closure_js_library")
+  deps += [
+      ctx.attr._grpc_web_abstractclientbase,
+      ctx.attr._grpc_web_clientreadablestream,
+      ctx.attr._grpc_web_error,
+      ctx.attr._grpc_web_grpcwebclientbase
+  ]
```
stanley-cheung: I wonder why only these 4 classes? Why not `Status` and `StatusCode` etc?

bazel/closure_grpc_web_library.bzl
```diff
@@ -0,0 +1,195 @@
...
+        ),
+        "_grpc_web_clientreadablestream": attr.label(
+            default = Label("//javascript/net/grpc/web:clientreadablestream"),
+        ),
+        "_grpc_web_error": attr.label(
+            default = Label("//javascript/net/grpc/web:error"),
+        ),
+        "_grpc_web_grpcwebclientbase": attr.label(
+            default = Label("//javascript/net/grpc/web:grpcwebclientbase"),
+        ),
```
stanley-cheung: Ditto, why just these 4 classes?

WORKSPACE
```diff
@@ -1,12 +1,13 @@
 workspace(name = "com_github_grpc_grpc_web")
 
+# TODO(yannic): Update to use official repository.
+# See https://github.com/bazelbuild/rules_closure/pull/278
 http_archive(
     name = "io_bazel_rules_closure",
-    sha256 = "a80acb69c63d5f6437b099c111480a4493bad4592015af2127a2f49fb7512d8d",
-    strip_prefix = "rules_closure-0.7.0",
+    sha256 = "248a7a98eb3962d9f0013e543ea79c5063a83bac7af349ebf412d844e6ab3035",
+    strip_prefix = "rules_closure-53f2cab21fa6c608f32f114387d88ffd7868c5fc",
```
stanley-cheung: Why this particular commit? Something new available after 0.7.0?

bazel/closure_grpc_web_library.bzl
Yannic: We only need to add the directly `goog.required` classes to deps, `Status`,  `StatusCode`, etc are included as transitive deps by those 4 classes.
https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpc_generator.cc#L409
(Only mode `GrpcWeb` seems to be open source).

bazel/closure_grpc_web_library.bzl
Yannic: See above

WORKSPACE
Yannic: This commit includes https://github.com/bazelbuild/rules_closure/pull/278, which is required for this because `closure_proto_aspect`, which we need to generate the normal proto js files, is not exposed by `rules_closure` (Everything which starts with `_` is private in bazel).

bazel/closure_grpc_web_library.bzl
```diff
@@ -0,0 +1,195 @@
...
+    attrs = {
+        'deps': attr.label_list(
+            mandatory = True,
+            providers = ['proto', 'closure_js_library'],
+            # The files generated by this aspect are required dependencies.
+            aspects = [closure_proto_aspect],
+        ),
+        'import_style': attr.string(
+            default = 'closure',
+            values = ['closure', 'commonjs'],
```
Yannic: I think I'd like to remove `commonjs` as a possible value for now. Most of `rules_closure` doesn't work with `commonjs` anyway.

scripts/kokoro.sh
```diff
@@ -25,6 +25,7 @@ docker build -t grpc-web:ubuntu_16_04 \
 docker-compose build
 
 bazel test javascript/net/grpc/web/...
+bazel test net/grpc/gateway/examples/...
```
stanley-cheung: Looks like this needs to be `bazel build` for now, instead of `bazel test`. Since there's no test target for now, the exit status of this command is 4, which invalidates the whole kokoro run.

bazel/closure_grpc_web_library.bzl
stanley-cheung: Is it because the output of this rule is expected to mostly feed into other `closure_js_library` rule, which expects `goog.require()` style includes anyways?

WORKSPACE
stanley-cheung: So we can only merge this PR after  bazelbuild/rules_closure#278 got merged right?

oferb: Technically you can merge this, you'll just be dependent on a commit that's not in `rules_closure`'s `master` branch (until it gets merged, that is).

bazel/closure_grpc_web_library.bzl
Yannic: The restriction to `goog.require()` is required because `closure_proto_{aspect,library}` (which generates the protobuf messages for us to avoid duplicate provide errors) does not support `commonjs` import style (yet). I will create a follow-up PR once this is fixed in rules_closure (Maybe we can create an issue for this and assign it to me).

WORKSPACE
Yannic: It's exactly like @oferb said, there is no technical reason to block this until the other PR is merged.

Yannic:
> 1. The output of this `closure_grpc_web_library` rule should be able to feed into other rules, like a closure_js_library.

Exactly. `closure_grpc_web_library` can be used, and is only really useful, as a dependency of `closure_js_{*}`.

> 2. We should see the <proto>.grpc.js output in the bazel-bin directory somewhere?

Well, the `<proto>.grpc.js` output may only be created if another target depends on it to reduce build time. This is similar to rules in rules_closure.

scripts/kokoro.sh
Yannic: Done

PR #222, opened 2018-08-06
Add documentation for closure_grpc_web_library

Yannic:


Yannic:
This is ready for review now.

README.md
```diff
@@ -105,6 +105,88 @@ here.
   - Only unary calls are supported for now.
 
 
+### Bazel
```
stanley-cheung: 1. Can we move this section between "TypeScript Support" and "Proxy Interoperability"
 2. Can we use two hashes headings "## Bazel Support"?

README.md
```diff
@@ -105,6 +105,88 @@ here.
...
+> :exclamation: ***Disclaimer***:
+> Using gRPC-Web in bazel currently requires [rules_closure] with
+  an unmerged patch \(See [#223](https://github.com/grpc/grpc-web/issues/223)\).
+
+Add the following to your WORKSPACE:
+
+```python
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
stanley-cheung: I see the intention here and I think it's a good idea. But this section seems a bit too heavy compared to the rest of the README. Can we shorten it somehow or take the details out to another page? Like perhaps we can move some of the details to https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/tutorial.md?

README.md
```diff
@@ -105,6 +105,88 @@ here.
...
+    "closure_grpc_web_library",
+)
+
+load(
+    "@io_bazel_rules_closure//closure:def.bzl",
+    "closure_js_library",
+    "closure_proto_library",
+)
+
+proto_library(
```
stanley-cheung: I think highlighting these 4 rules is important. Instead of just putting all 4 rules as codes, can we explain what each of these 4 rules are for and why do we need 4 of them?

sampajano:
Hi.. I'm closing this PR because all the JS Bazel rules are cleaned up in https://github.com/grpc/grpc-web/pull/1138..

So this change no longer applies.. :)

Sorry for the change!

PR #240, opened 2018-08-10
webpack demo

### LINKED ISSUE #98: Node deployment and webpack support
Redirected from https://github.com/grpc/grpc/issues/11435

@slootzky
@murgatroid99  (you may want to close the original issue)

stanley-cheung:
This is to show that the grpc-web and protobuf generated code can be compiled with webpack.

Fixes #98


benitogf:
hello @stanley-cheung think that it would be nice to have a separated repo with the demo

PR #246, opened 2018-08-14
Closure compiler download link uses plain HTTP

wfhartford:
I may be wrong, but shouldn't this download happen over HTTPS?


stanley-cheung:
Thanks for your contribution!

PR #249, opened 2018-08-17
Various fixes to codegen plugin

### LINKED ISSUE #244: CommonJS require paths when proto files are in subdirectories
Lets say we have `.proto` files organised into directories:
```
apis/myproject/myapi/v1/myapi.proto
apis/myproject/otherapi/v1/otherapi.proto
```
Generating code involves including the `apis` directory:
```
$ protoc -I./apis  ...  ./apis/myproject/myapi/v1/myapi.proto
```
The generated code might be output to:
```
src/myproject/myapi/v1/
src/myproject/otherapi/v1/
```
The `google/protobuf` and `grpc/grpc` plug-ins generated `require` calls like this:
```
... = require("../../../myproject/myapi/v1/myapi_pb.js");
```
However, the `grpc-web` plug-in generates `require` calls like this:
```
... = require('./myproject/myapi/v1/myapi_pb.js');
```
The generated `require` paths should account for the directory structure relative to the import path.

### LINKED ISSUE #245: Support compiling multiple .proto files with a single command
`protoc` supports passing multiple `.proto` files with a single call. Typically, the output directory is specified and the file names are based on the names of the `.proto` files.

Currently, if more than one `service` is defined in more than one `.proto` file, `protoc` returns an error:

```
... Tried to write the same file twice.
```

stanley-cheung:
Fixes #244 
Fixes #245 

This PR introduces the following changes:
 - Allow source proto files, specified in subdirectories, to be output'ed to the corresponding subdirectories, relative to the base output directory (#244)
 - Allow multiple source proto files, to be generated into its own `<proto>_grpc_pb.js` files (#245)
 - Fixed an issue where, if any of the request or response type within a service definition points to a type specified in another file (or package), the generated code was not properly alias'ed.
 - Added tests for all the above scenarios
 - Added tests for closure-style generated code

cc @rwlincoln


stanley-cheung:
`out=` should no longer be required. When not specified, filename should be in the form of `<proto>_grpc_pb.js`

rwlincoln:
I checked out the correct commit and the `out=` option is indeed no longer required. The `require` paths account for the directory hierarchy and it is possible to pass multiple `.protos` into a single `protoc` call.

Can the generated files be named `<proto>_grpcweb_pb.js` or something similar? This would prevent files generated by the gRPC Node.js plug-in from being overwritten.

stanley-cheung:
Sounds good. Generated files are now named `<proto>_grpc_web_pb.js`

Issue #255, opened 2018-08-21
Add Typescript definition file for the grpc-web npm package


aberasarte:
Now that the grpcweb runtime is available as an npm package 🎉 🎉 , it would be even better if the package could include a Typescript definition file as well. This way, it would be much easier to consume the grpcweb runtime from Typescript code. 

If this isn't feasible I'm thinking about submitting a d.ts file to [definitely typed](https://github.com/DefinitelyTyped/DefinitelyTyped) but this is not the best solution.

jonahbron:
I might be able to contribute a new file with the basic outline this week.

glerchundi:
This would be fantastic!

stanley-cheung:
I started a PR #258 to show how the `index.d.ts` might work. Any feedback appreciated!

stanley-cheung:
PR #258 merged. The `index.d.ts` typings should be making it into the next release `0.4.0` currently planned for next week. 

After that user should be able to do `import * as grpcWeb from 'grpc-web';` after an `npm i grpc-web`.

aberasarte:
Great! Thanks for adding this @stanley-cheung .

stanley-cheung:
The npm package version `0.4.0` has been pushed. Please try it out!

PR #258, opened 2018-08-22
Experimental Typescript support

stanley-cheung:
For #255 

I am going to try to expand on this. For my very initial understanding, this seems to work.

Feedback highly appreciated!

cc: @jonahbron

### DIFF
--- a/packages/grpc-web/index.d.ts
+++ b/packages/grpc-web/index.d.ts
@@ -1,0 +1,73 @@
+declare module "grpc-web" {
+
+  export interface Metadata { [s: string]: string; }
+
+  export namespace AbstractClientBase {
+    class MethodInfo {
+      constructor (responseType: {},
+                   requestSerializeFn: (request: {}) => {},
+                   responseDeserializeFn: (bytes: {}) => {});
+    }
+  }
+
+  export class AbstractClientBase {
+    rpcCall (method: string,
+             request: {},
+             metadata: Metadata,
+             methodInfo: AbstractClientBase.MethodInfo,
+             callback: (err: Error, response: {}) => void
+            ): ClientReadableStream;
+
+    serverStreaming (method: string,
+                     request: {},
+                     metadata: Metadata,
+                     methodInfo: AbstractClientBase.MethodInfo
+                    ): ClientReadableStream;
+  }
+
+  export class ClientReadableStream {
+    on (type: string,
+        callback: (...args: Array<{}>) => void): ClientReadableStream;
+    cancel (): void;
+  }
+
+  export interface GrpcWebClientBaseOptions {
+    format?: string;
+    suppressCorsPreflight?: boolean;
+  }
+
+  export class GrpcWebClientBase extends AbstractClientBase {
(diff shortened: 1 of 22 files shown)


stanley-cheung:
So far, this PR demos how a typescript project can make use of the `index.d.ts` typings for the `grpc-web` module. I hand-wrote an `echo_grpc_web_pb.ts` file that could be generated from a future grpc-web x TS protoc plugin. To test whether the `index.d.ts` typings do matter, I flip, let's say, a `hostname` parameter from a `string` to a `number`. Now the `tsc` command will fail because a string is being passed.

jonahbron:
That looks right to me.  One alternative to the `echo_grpc_web_pb.ts` file I've seen in [ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen) is producing .d.ts for the generated files too.  That way the JS generator can still be used as-is.

packages/grpc-web/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  export class ClientReadableStream {
+    on (type: string,
+        callback: (...args: any[]) => any): ClientReadableStream;
+    cancel (): void;
+  }
+
+  export interface GrpcWebClientBaseOptions {
+    format?: string;
```
aberasarte: I think that we could add the `suppressCorsPreflight` option as well: `suppressCorsPreflight?: boolean`

stanley-cheung: Will do.

stanley-cheung:
> That looks right to me. One alternative to the echo_grpc_web_pb.ts file I've seen in ts-protoc-gen is producing .d.ts for the generated files too. That way the JS generator can still be used as-is.

Will try that. Thanks.

stanley-cheung:
Added a full Typescript example up to this point.

`ts-example/EchoServiceClient.ts`: a hand-written would-be full-TS codegen output

`ts-example/echo_grpc_web_pb.d.ts`: .d.ts typings that is meant to work alongside existing `echo_grpc_web_pb.js`  CommonJS codegen output
`ts-example/echo_pb.d.ts`: .d.ts typings for protobuf message types, meant to work alongside `echo_pb.js` like above

`ts-example/client.ts`: TS client app code

Uncomment either Line 25 or Line 26 to try between the two modes above:
```
import {EchoServiceClient} from './echo_grpc_web_pb';   // .d.ts typings for existing JS codegen output
// import {EchoServiceClient} from './EchoServiceClient';  // full-TS codegen output
```

stanley-cheung:
Added 2 modes to the `protoc-gen-grpc-web` protoc plugin

 - `import_style=commonjs+dts`: this will generate `.d.ts` typings for both the `<proto>_grpc_web_pb.js` service stub and the `<proto>_pb.js` proto message classes
 - `import_style=typescript`: this will generate `<proto>ServiceClientPb.ts` typescript service stub

stanley-cheung:
Squashing commits. Got internal reviewed. Will merge this soon.

PR #259, opened 2018-08-22
Add "strictDependencies" to suppress attribute of Bazel rule

Yannic:
Not suppressing this causes an error for a missing direct dependency
for the generated gRPC-Web library if a service returns a protobuf
message not declared in the same proto_library.


stanley-cheung:
@Yannic this PR looks good. Can you rebase this? I also left some comments on the other outstanding PR, please take a look. Thanks for your contributions!

Yannic:
@stanley-cheung Thanks for your review(s)! Rebased #259 and #222, will take fix the failing tests in #253 soon.

Issue #307, opened 2018-10-04
Cannot access options


mitar:
In our proto files we do something like:

```
extend google.protobuf.FileOptions {
    string protocol_version = 54001;
}

option (protocol_version) = "1.2.3";
```

This allows us to encode the version of the API into the API itself. Then user of the API can do (in Python):

```python
version = core_pb2.DESCRIPTOR.GetOptions().Extensions[core_pb2.protocol_version]
```

To retrieve this constant from the API file.

I see in generated JavaScript file that file options are extended. But the option itself is not stored anywhere. Could you add it as well?

stanley-cheung:
This may be an issue for [protobuf](https://github.com/protocolbuffers/protobuf), but I will confirm.

stanley-cheung:
Yea this is a protobuf issue: https://github.com/protocolbuffers/protobuf/issues/3554

mitar:
Oh, my own issue there. How time flies. :-)

stanley-cheung:
Yea that part was controlled by `--js_out`. This repo's plugin is responsible for `--grpc-web_out`.

PR #311, opened 2018-10-06
Document how to prevent Envoy to timeout streaming

### LINKED ISSUE #309: If stream messages take long, streaming disconnects
I am using Envoy proxy with Python server-side and i am trying to stream messages. I discovered that if messages take a long time to arrive, an error is printed in the browser console and no more events for that stream come. The error is:

```
POST http://localhost:8080/Core/GetResults net::ERR_INCOMPLETE_CHUNKED_ENCODING 200 (OK)
```

I am guessing that Envoy proxy disconnects the connection. It seems it does that at exactly 15 seconds (that's what I see in Chrome web inspector network tab).

mitar:
Fixes #309.

The example is missing this option, so all streaming calls get terminated after 15 seconds, if there is no message sent.

### DIFF
--- a/net/grpc/gateway/examples/helloworld/README.md
+++ b/net/grpc/gateway/examples/helloworld/README.md
@@ -112,7 +112,9 @@ static_resources:
               domains: ["*"]
               routes:
               - match: { prefix: "/" }
-                route: { cluster: greeter_service }
+                route:
+                  cluster: greeter_service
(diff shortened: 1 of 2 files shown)


mitar:
So now just that #307 is fixed and I will have a fully interoperable client in browser. It works really cool otherwise. Thanks for this.

PR #316, opened 2018-10-06
Add test for PromiseClient

stanley-cheung:
Add a unit test to exercise the PromiseClient generated code.

cc @mitar @juanjoDiaz

### DIFF
--- a/packages/grpc-web/test/plugin_test.js
+++ b/packages/grpc-web/test/plugin_test.js
@@ -56,8 +56,8 @@ describe('grpc-web plugin test, with subdirectories', function() {
   beforeEach(function() {
     removeDirectory(path.resolve(__dirname, GENERATED_CODE_PATH));
     fs.mkdirSync(path.resolve(__dirname, GENERATED_CODE_PATH));
-    MockXMLHttpRequest = mockXmlHttpRequest.newMockXhr()
-    global.XMLHttpRequest = MockXMLHttpRequest; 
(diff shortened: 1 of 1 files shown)


packages/grpc-web/test/plugin_test.js
```diff
@@ -214,4 +216,38 @@ describe('grpc-web plugin test, proto with no package', function() {
...
+    const {GreeterPromiseClient} = require(genCodePath2);
+    var myClient = new GreeterPromiseClient("MyHostname", null, null);
+    assert.equal('function', typeof myClient.sayHello);
+
+    var p = myClient.sayHello(new HelloRequest(), {});
+    assert.equal('function', typeof p.then);
+  });
+
+  it('PromiseClient: response should resolve promise', function() {
+    execSync(genCodeCmd);
```
juanjoDiaz: Is it needed to run this on each test method or could it be run in the `beforeAll` method?

Same applies to`HelloRequest`, `GreeterPromiseClient`, etc.

packages/grpc-web/test/plugin_test.js
```diff
@@ -214,4 +216,38 @@ describe('grpc-web plugin test, proto with no package', function() {
...
+
+    const {HelloRequest} = require(genCodePath1);
+
+    const {GreeterPromiseClient} = require(genCodePath2);
+    var myClient = new GreeterPromiseClient("MyHostname", null, null);
```
juanjoDiaz: More of a style comment but do you have any standard on when to use const/let/var?

PR #324, opened 2018-10-13
Internal sync

stanley-cheung:
- removed some trailing whitespace
 - some lines longer than 80 characters
 - missing a semicolon somewhere
 - some JSDoc is required
 - I have to add back some deleted code because turns out some example was shared with some internal shared demo and I can't just delete them.

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -16,12 +16,12 @@
  *
  */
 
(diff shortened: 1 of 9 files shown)


hunnain:
@stanley-cheung  please also review my code i have also created an React Example please merge my code also

stanley-cheung:
@hunnain Please don't abuse the issue tracking system. This PR is unrelated to yours.

Issue #279, opened 2018-09-03
Typescript Type Check Errors


purebluesong:
Some errors exposed when I try to compile my generated typescript code.

One error message here:
```
Argument of type '(request: Empty) => Uint8Array' is not assignable to parameter of type '(request: {}) => {}'.
  Types of parameters 'request' and 'request' are incompatible.
    Type '{}' is not assignable to type 'Empty'.
      Property 'serializeBinary' is missing in type '{}'.
```

And I found the generation code is here:
https://github.com/grpc/grpc-web/blob/86631990e365130d5cfca3aa93613ec1eeef7da4/javascript/net/grpc/web/grpc_generator.cc#L395

Another error message here:
```
Argument of type '(err: Error, response: Empty) => void' is not assignable to parameter of type '(err: Error, response: {}) => void'.
  Types of parameters 'response' and 'response' are incompatible.
    Type '{}' is not assignable to type 'Empty'.
```

And I found the generation code is here:
https://github.com/grpc/grpc-web/blob/86631990e365130d5cfca3aa93613ec1eeef7da4/javascript/net/grpc/web/grpc_generator.cc#L438

The part of the wrong generated code shows here:
```
methodInfoSome = new grpcWeb.AbstractClientBase.MethodInfo(
    Empty,
    (request: SomeRequest) => {  // Here is the first error
      return request.serializeBinary();
    },
    Empty.deserializeBinary
  );

  some(
    request: SomeRequest,
    metadata: grpcWeb.Metadata,
    callback: (err: grpcWeb.Error,
               response: Empty) => void) {
    return this.client_.rpcCall(
      this.hostname_ +
        '/SomePath',
      request,
      metadata,
      this.methodInfoSome,
      callback); // Here is the second error which the callback is `(err: Error, response: Empty) `
  }
```

The `Empty` message defined in proto file likes this `message Empty {}`
FYI: My TS version is 3.0.1

stanley-cheung:
I am not quite able to reproduce. Can you show me your `.proto` file, preferably a little simplified to illustrate the error, and perhaps an entry point `client.ts`?

purebluesong:
The proto is here:
```
service SomeService {
    rpc Some(SomeRequest) returns (Empty) {}
}

message SomeRequest {
    string id = 1;
}

message Empty {}
```

And the entry point is nothing special just a useless import.

stanley-cheung:
Hm, I am still missing something.

[...]

PR #345, opened 2018-10-27
update typescript generation to work in strict mode

henriiik:
The changes in this PR fixes the issues brought up in #279.

The problem is that in typescripts strict mode, the type `{}` only matches empty objects ex `const someVar = {}`, and not anything else.

The following changes are made:

1. update `tsconfig.json` to use `strict`
2. update `MethodInfo`/`AbstractClientBase`/`ClientReadableStream` to be generic over `Request` and `Response` types
3. update `ClientReadableStream.on` to have correct `callback` type depending on `type` argument
4. update generated clients to allow passing of `null` as `credentials` and `options`

I could not figure out where `credentials` and `options` are used so I could not figure out the correct types. If you could point me in the right direction that would be great! :)


henriiik:
Thanks for the info @stanley-cheung 🙂

I have updated the PR with the correct types for `credentials` and `options` and it should be good to merge now.

PR #352, opened 2018-11-06
 Add a max grpc timeout to the echo example. 

mjduijn:
When a beginner continues with the example code to set up a streaming service, streams will be interrupted for no apparant reason after ~14 seconds. 
This makes sure streams will stay open.


stanley-cheung:
Reference: #311

PR #365, opened 2018-11-10
Fix response header value with colon

### LINKED ISSUE #318: Error message with colon does gets changed on the way to the client app
I have a server-side implemented in GRPC and I throw there an exception with message like "Foo: bar". This seems to be sent to the client with header:

```
grpc-status: 3
grpc-message: Foo: bar
```

But in the JavaScript callback on the client I get only `{code: "3", message: "Foo"}`.

I think the issue is that because of `:` character there, header parsing seems to break or something. I tried to find the code where this breaks but I was not yet successful.

stanley-cheung:
Fixes #318.

Very strange, let's say we have this response header `grpc-message: Foo: Bar` sent back from gRPC backend service, and thru Envoy, from the browser console, the header value apparently was sent back properly.

But, if I call `xhr.getResponseHeaders()` to get the whole bag of response headers, one of the key:value pair was `grpc-message: Foo`.

OTOH, if I call `xhr.getResponseHeader('grpc-message')`, I get `Foo: Bar` back properly.

I couldn't find a bug from Chrome or Closure to report this strange behavior with `xhr.getResponseHeaders()`.

Also OTOH, I cannot find any HTTP spec that said that http header value cannot have a colon in it. So apparently it's OK to have header value that has a colon in it.

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientreadablestream.js
+++ b/javascript/net/grpc/web/grpcwebclientreadablestream.js
@@ -195,10 +195,10 @@ const GrpcWebClientReadableStream = function(genericTransportInterface) {
     }
     var responseHeaders = self.xhr_.getResponseHeaders();
     if (GRPC_STATUS in responseHeaders &&
-        responseHeaders[GRPC_STATUS] != StatusCode.OK) {
+        Number(self.xhr_.getResponseHeader(GRPC_STATUS)) != StatusCode.OK) {
       self.onErrorCallback_({
-        code: Number(responseHeaders[GRPC_STATUS]),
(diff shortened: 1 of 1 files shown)


javascript/net/grpc/web/grpcwebclientreadablestream.js
```diff
@@ -195,10 +195,10 @@ const GrpcWebClientReadableStream = function(genericTransportInterface) {
     }
     var responseHeaders = self.xhr_.getResponseHeaders();
```
mitar: Why even use `getResponseHeaders` at all?

PR #389, opened 2018-11-27
Make metadata and callback optional

shaxbee:


shaxbee:
Optional as in it can be nullable. This is consistent with nodejs grpc package API.

Only change in `abstractclientbase.js` is this:
```
if (metadata) {
  xhr.headers.addAll(metadata);
}
```

PR #394, opened 2018-11-28
Bazel rules: additional suppress options

factuno-db:
I had to add these suppress options in order to get a real-life generated service.grpc.js to compile with the Closure compiler.


factuno-db:
Sure thing! I just opened #395 . Maybe you could also help me out with https://github.com/bazelbuild/rules_closure/pull/314 since you already approved https://github.com/bazelbuild/rules_closure/pull/277 ;-)

Yannic:
Thanks! I'll take a look at bazelbuild/rules_closure#314 when I find some time, but we're gonna have to bug Dimitry to actually merge it.

PR #404, opened 2018-12-03
Generate Typescript definition for top level Enums

### LINKED ISSUE #403: TS declarations files are missing enums
Hello 👋!

I noticed that `Enum` types are missing in the generated `.d.ts` files. This is an example

```.go
syntax = "proto3";

enum Examples {
  CASE_1 = 0;
  CASE_2 = 1;
  CASE_3 = 2;
  CASE_4 = 3;
  CASE_5 = 4;
}
```

This file will generated the JS but not the declaration file
```js
/**
 * @fileoverview
 * @enhanceable
 * @suppress {messageConventions} JS Compiler reports an error if a variable or
 *     field starts with 'MSG_' and isn't a translatable message.
 * @public
 */
// GENERATED CODE -- DO NOT EDIT!

var jspb = require('google-protobuf');
var goog = jspb;
var global = Function('return this')();

goog.exportSymbol('proto.Examples', null, global);
/**
 * @enum {number}
 */
proto.Examples = {
  CASE_1: 0,
  CASE_2: 1,
  CASE_3: 2,
  CASE_4: 3,
  CASE_5: 4
};

goog.object.extend(exports, proto);
```

I made sure to run the command with the argument `--grpc-web_out=import_style=commonjs+dts`

Thanks in advance

rogchap:
Fixes #403

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -675,6 +675,10 @@ void PrintProtoDtsFile(Printer *printer, const FileDescriptor *file)
        it != messages.end(); it++) {
     PrintProtoDtsMessage(printer, it->second, "");
   }
+  
+  for (int enum_index = 0; enum_index < file->enum_type_count(); ++enum_index) {
+    PrintProtoDtsEnum(printer, file->enum_type(enum_index));
+  }
 }
 
 void PrintFileHeader(Printer* printer, const std::map<string, string>& vars) {


rogchap:
This only works if the `enum` is declared within a service definition file. 
Looking to see how we get a TypeScript definition of a shared `enum`, ie. In its own file: `myenum.proto` and then imported.

stanley-cheung:
Thanks for the great contributions! Yea looks like this is a gap that we need to address.

On the other hand, I'd like to start requesting contributors to also add tests for features they added. We already have some tests - running `npm test` from the `packages/grpc-web` directory will trigger those. 

This will fit in to something like `https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/plugin_test.js`. In the test, we assume the `protoc-gen-grpc-web` is already installed somewhere in the system (in reality, before `npm test` is run, we will re-compile the plugin every time a PR is updated), then we run the `protoc ... --grpc-web_out=...` command to actually generate the client stub, load the file via a `require()` statement and do some test on the client stub class. Let's do something similar and add a test for this Enum feature.

Thanks!

rogchap:
@stanley-cheung Totally agree that we should have tests for this.
Hard to load the file via `require()` as the JS is untouched and it's the TypeScript generation that we need to test.
Do you have any advise on how to test the TypeScript generation; my best thought is doing a string comparison/regex on the file, but that seems ugly.

stanley-cheung:
I am also hoping someone from the community can give us some advice on how's best to test TS code generation.

Quick thought: perhaps we can `exec('tsc')`?  At least that should weed out some TS compilation errors. I am actually not sure what's the "standard" way of testing TS code in a npm package these days. Also kinda new to this :)

johanbrandhorst:
Looking at the TS protoc generator Improbable has written, it seems they generate the files and just check properties on the generated types: https://github.com/improbable-eng/ts-protoc-gen/blob/master/test/integration/service/grpcweb.ts.

rogchap:
So... I've added a basic test to make sure the TypeScript can compile. I manually tested `tsc` to see if it would fails on a malformed typescript definition file, which it does; so at least we know the TypeScript is right as far as it can compile.

The thing is, we can't test any output from the TypeScript to validate as this is only a definition file and the actual JS is already generated 💁‍♂️.

pacurtin:
Awesome work you guys are doing. I'm currently using JS web-grpc but I'm looking forward to switching to TS. I'm waiting for this fix to be released so I can do so.

Sorry if this is the wrong place to ask but do you know when the next release is scheduled for?

PR #413, opened 2018-12-18
Emit status event on empty stream response

shaxbee:
When stream is closed immediately by the server grpc-status and grpc-message headers are populated. 
Current implementation is broken as it does not fire `status` or `end` event in that case.


shaxbee:
@stanley-cheung Done! :-)

shaxbee:
Fixed issues reported by kokoro.

stanley-cheung:
@shaxbee There seems to be still some error reported by Kokoro. Please take a look. Thanks!

shaxbee:
@stanley-cheung Sorry, missed that one :-)

PR #427, opened 2019-01-03
Promise function should use ES5 functions rather than fat arrows (IE browser support)

rogchap:
Arrow functions are not compatible with IE11, this fix allows the generated JS to be used directly in the browser without going through a seperate transpiler (Babel)


rogchap:
@stanley-cheung Keen to get your thoughts on this one; You back from the holidays?

stanley-cheung:
There are other places we are spitting out arrow functions but those are for TypeScript output so we should be fine there. This is the only place for the normal JS output.

rogchap:
Thanks @stanley-cheung

PR #433, opened 2019-01-10
Match name nesting and imports in .d.ts with .js files

shaxbee:
Exported names were incorrectly mangled and didn't reflect types in `*
pb.js` files, resulting in runtime errors where trying to instantiate.
Type aliases were replaced with proper nesting based on [improbable-eng/ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen).
Aliased imports reflecting `*pb.js` files were added.
Services now return correct message types for external protos.


stanley-cheung:
@shaxbee Thanks for this contributions. Can you give us an example on what this achieves? Like maybe a couple of simple .protos and what the current and new behavior would be? Thanks

PR #441, opened 2019-01-22
Correct the link to PROTOCOL-WEB - it's in the grpc repo

takamori:
It's in the grpc/grpc repo rather than the grpc/grpc-web repo


stanley-cheung:
Dupe of #440 Thanks for the fix anyways!

PR #444, opened 2019-01-25
feature: Messages in typings extending jspb.Message

### LINKED ISSUE #442: [feature-request] Include "extends Message" on generated classes in .d.ts files
Hello 👋

The classes defined in the generated `*_pb.td.s` files are missing the statement `extends Message`  despite they do inherit the method from `Message` class from `google-protobuf`.

Because of this, TypeScript doesn't recognize inherited methods like `clone` and these can be quite handy (for cloning an instance, for example)

shaxbee:
In generated protobuf typings extend jspb.Message to expose methods, add missing parameters and static methods.

Example:
https://github.com/shaxbee/grpc-web-dts-example/blob/master/typings-extend-jspb-message/example/example_pb.d.ts

Resolves #442

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -677,7 +677,7 @@ void PrintProtoDtsMessage(Printer *printer, const Descriptor *desc, const FileDe
   std::map<string, string> vars;
   vars["class_name"] = class_name;
 
-  printer->Print(vars, "export class $class_name$ {\n");
+  printer->Print(vars, "export class $class_name$ extends jspb.Message {\n");
   printer->Indent();
   printer->Print("constructor ();\n");
   for (int i = 0; i < desc->field_count(); i++)
@@ -688,9 +688,12 @@ void PrintProtoDtsMessage(Printer *printer, const Descriptor *desc, const FileDe
     printer->Print(vars, "set$js_field_name$(a: $js_field_type$): void;\n");
   }
   printer->Print(vars,
-                "toObject(): $class_name$.AsObject;\n"
                 "serializeBinary(): Uint8Array;\n"
-                "static deserializeBinary: (bytes: {}) => $class_name$;\n");
+                "toObject(includeInstance?: boolean): $class_name$.AsObject;\n"
+                "static toObject(includeInstance: boolean, msg: $class_name$): $class_name$.AsObject;\n"
+                "static serializeBinaryToWriter(message: $class_name$, writer: jspb.BinaryWriter): void;\n"
+                "static deserializeBinary(bytes: Uint8Array): $class_name$;\n"
+                "static deserializeBiinaryFromReader(message: $class_name$, reader: jspb.BinaryReader): $class_name$;\n");
   printer->Outdent();
   printer->Print("}\n\n");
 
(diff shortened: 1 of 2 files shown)


jmzwcn:
great!

javascript/net/grpc/web/grpc_generator.cc
```diff
@@ -723,6 +726,7 @@ void PrintProtoDtsMessage(Printer *printer, const Descriptor *desc, const FileDe
 
 void PrintProtoDtsFile(Printer *printer, const FileDescriptor *file)
 {
+  printer->Print("import * as jspb from \"google-protobuf\"\n\n");
```
stanley-cheung: Test failure:
```
/github/grpc-web/net/grpc/gateway/examples/echo/ts-example/node_modules/grpc-web -> /usr/local/lib/node_modules/grpc-web -> /github/grpc-web/packages/grpc-web
echo_pb.d.ts(1,23): error TS7016: Could not find a declaration file for module 'google-protobuf'. '/github/grpc-web/net/grpc/gateway/examples/echo/ts-example/node_modules/google-protobuf/google-protobuf.js' implicitly has an 'any' type.
  Try `npm install @types/google-protobuf` if it exists or add a new declaration (.d.ts) file containing `declare module 'google-protobuf';`
Service 'ts-client' failed to build: The command '/bin/sh -c cd $EXAMPLE_DIR/ts-example &&   npm install &&   npm link grpc-web &&   tsc &&   npx webpack &&   cp echotest.html /var/www/html &&   cp dist/main.js /var/www/html/dist' returned a non-zero code: 2
```

The problem with this is that, the `google-protobuf` package does not publish a `.d.ts` typings.

stanley-cheung: adding `@types/google-protobuf` to `package.json` might help?

shaxbee:
@stanley-cheung Added `@types/google-protobuf` and checked with `docker-compose up --build` that it builds now.

javascript/net/grpc/web/grpc_generator.cc
```diff
@@ -688,9 +688,12 @@ void PrintProtoDtsMessage(Printer *printer, const Descriptor *desc, const FileDe
...
   }
   printer->Print(vars,
-                "toObject(): $class_name$.AsObject;\n"
                 "serializeBinary(): Uint8Array;\n"
-                "static deserializeBinary: (bytes: {}) => $class_name$;\n");
+                "toObject(includeInstance?: boolean): $class_name$.AsObject;\n"
+                "static toObject(includeInstance: boolean, msg: $class_name$): $class_name$.AsObject;\n"
+                "static serializeBinaryToWriter(message: $class_name$, writer: jspb.BinaryWriter): void;\n"
+                "static deserializeBinary(bytes: Uint8Array): $class_name$;\n"
+                "static deserializeBiinaryFromReader(message: $class_name$, reader: jspb.BinaryReader): $class_name$;\n");
```
rogchap: is that a typo?

```diff
- static deserializeBiinaryFromReader
+ static deserializeBinaryFromReader

shaxbee: Looks like it, could you make PR?

stanley-cheung: I fixed it directly on master. Thanks for the catch.

PR #400, opened 2018-11-30
Included support for JSType maps. 

thomasamills:
The generated code was previously setting the type as a repeated message.

PR #452, opened 2019-02-03
fix: exclude map entry message from typings, fix optional values

shaxbee:
This PR brings bindings in line with `ts-protoc-gen`.

Do not generate typings for proto `map` entry type.
Fix optionality of `repeated` fields - array fields are required.
Add `has$field_name$` method for message fields.
Add `add$element_name$` method for repeated fields.
Support `[jstype=JS_STRING]` field option.
Prefix reserved fields in AsObject with `pb_` prefix.

Example:
https://github.com/shaxbee/grpc-web-dts-example/blob/master/map-asobject/example/example_pb.d.ts

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -218,19 +218,18 @@ string ModuleAlias(const string& filename) {
 }
 
 string JSMessageType(const Descriptor *desc, const FileDescriptor *file) {
-    string result;
-    if (desc->file() != file) {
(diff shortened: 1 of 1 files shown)


shaxbee:
This and #448 supersedes #400

stanley-cheung:
Sorry for the late look at this. There had been some work to re-sync this repo with our internal codebase so it looks like there are some new conflicts.

Can you please resolve them and rebase to the latest master please? Thanks

shaxbee:
@stanley-cheung Rebased :-)

Issue #467, opened 2019-02-20
grpc-web client 'end' or 'status' callbacks never called


subhan-nadeem:
I have a Java GRPC server with SSL enabled. Every time I make a stream request with my backend, the 'end' and 'status' callbacks are never triggered. The 'data' callback works just fine.

Pinging the same API with a Node.js or Java GRPC client results in a callback to the 'end' callback.

Server is on Linux, client is Windows + Chrome.

Similarly for the helloworld example, if I set 'end' and 'status' callbacks, only the 'status' callback is triggered.

My Envoy configuration is as follows:

```
admin:
  access_log_path: /tmp/admin_access.log
  address:
    socket_address:
      protocol: TCP
      address: 127.0.0.1
      port_value: 9901
static_resources:
  listeners:
    - name: listener_0
      address:
        socket_address:
          protocol: TCP
          address: 0.0.0.0
          port_value: 12340
      filter_chains:
        - filters:
            - name: envoy.http_connection_manager
              config:
                stat_prefix: ingress_http
                stream_idle_timeout: 0s
                route_config:
                  name: local_route
                  virtual_hosts:
                    - name: local_service
                      domains: ["*"]
                      routes:
                        - match:
                            prefix: "/"
                          route:
                            host_rewrite: <host>
                            cluster: service
                            max_grpc_timeout: 60s
                      cors:
                        allow_origin:
                          - "*"
                        allow_methods: GET, PUT, DELETE, POST, OPTIONS
                        allow_headers: keep-alive,user-agent,cache-control,content-type,content-transfer-encoding,custom-header-1,x-accept-content-transfer-encoding,x-accept-response-streaming,x-user-agent,x-grpc-web,grpc-timeout
                        max_age: "1728000"
                        expose_headers: custom-header-1,grpc-status,grpc-message
                        enabled: true
                http_filters:
                  - name: envoy.grpc_web
                  - name: envoy.cors
                  - name: envoy.router
  clusters:
    - name: service
      connect_timeout: 0.25s
      type: LOGICAL_DNS
      # Comment out the following line to test on v6 networks
      dns_lookup_family: V4_ONLY
      lb_policy: ROUND_ROBIN
      hosts:
        - socket_address:
            address: <host>
            port_value: 443
      tls_context:
        sni: <host>
```

stanley-cheung:
Similar to #289 I believe

jonahbron:
[...]

Issue #384, opened 2018-11-24
Help with implementing Stream re-connection for grpc-web. 


weilip1803:
Currently, if the grpc service is not active or running, for example if echo service is not available. This only returns a post error request which does not goes into the stream.on('error', ...) callback unlike the nodejs api. Are there plans to make this consistent with the nodejs api? And is waitForReady() going to be implemented also?
```
var stream = echoService.serverStreamingEcho(streamRequest, metadata);
stream.on('error', function(err) {
  console.log(err);
});
```
Even for unary calls, the error does not trigger the call back function. Is that anyway to detect whether the service behind envoy is down a not?

Please let me know if there is another way for me to do graceful re-connection. Thanks in advance

EDIT:
I also noticed something funny about envoy. 
```
virtual_hosts:
            - name: local_service
              domains: ["*"]
              routes:
              - match: { prefix: "/a/" }
                route: { cluster: to_a, prefix_rewrite: "/", timeout: 0s }
              - match: { prefix: "/b/" }
                route: { cluster: to_b, prefix_rewrite: "/", timeout: 0s }
              - match: { prefix: "/c/" }
                route: { cluster: to_c, prefix_rewrite: "/", timeout: 0s }
```
Once one of the services disconnect, and I try to connect to another service which is available, envoy dosent seem to work anymore. 
1) Start up all service then envoy
2) take down service a.
3) Try to connect to service a with the client
4) Everything becomes not working till i restart envoy

pumano:
I got the same problem, also need clarification about it. 
I don't receive any error in `stream.on('error', ...)`

weilip1803:
I have just tried your example. You can try typing 1000 hello into the chat box and press send. Then you just need to kill the service while the node.js grpc server is still streaming the message. You will realize you will not get back the error code message 99% of the time.

weilip1803:
Okay i realize my mistake on my side it was a dump mistake. You can close this. But is examples/echo/echotest.html used at all? I think you should remove it it really confused me I spend 1 hour thinking that the dockerfile used that file. But other than that thanks for the help.

PR #497, opened 2019-03-14
Remove a return that skip emission of end callback

### LINKED ISSUE #289: `end` event occasionally not emitted
I am making a streaming RPC call and listening using `on()` to `data`, `status` and `end` events. I'm always getting the data, and always getting a status with code 0, but the `end` event only appears about 20% of the time. My client code only sends that one single streaming request.

Is this a known issue, or are there any obvious things I might be missing? If not, I'll share some minimal code to demonstrate this issue.

tinou98:
When the connection is closed and there isn't any message sent, the end callback was not called.

Should fix #289, #384 and #467


stanley-cheung:
@tinou98 Thanks for your contributions. Can you please rebase to the latest master? Thanks.

tinou98:
Rebased to latest master.

Issue #351, opened 2018-11-05
Documentation for Credentials


jakepoz:
We would like to be able to pass in CallCredentials to Grpc-web requests. I see that the Client objects support taking a credentials argument:

`  /**
   * @private @const {?Object} The credentials to be used to connect
   *    to the server
   */
  this.credentials_ = credentials;`

But I don't see that this is used anywhere. We're looking to add in a metadata Authorization token automatically on every request for example.

What is the proper way to implement this?

stanley-cheung:
You can add your Authorization token as a key:value in the `metadata` parameter per generated RPC method (usually the 2nd parameter). 

The `credentials` object in the client constructor is currently not hooked up yet.

jakepoz:
We are passing it now this way, but it requires remembering to do this per each call of the RPC. Is there a timeline for the credentials capability?

We have also considered another option, of simply sending cookies with each request and authenticating that way. Would this be reasonable?

johanbrandhorst:
Automatically attaching a header on outgoing requests is a typical use of a gRPC client interceptor (like http middleware). I think interceptor support would help solve this issue.

stanley-cheung:
Allowing sending cookies is tracked at #176 
But yes, an even better design is Interceptor support, which is tracked at #283.
Both are very much in our radar.

jakepoz:
Thank you Stanley, our solution for today is to structure the service on the same origin, so we can send cookies with the request automatically.

swuecho:
@jakepoz could you elaborate how to send the cookies with request direct without add metadata per rpc call? Thanks

jakepoz:
Yes, our general stack is a React Frontend (made with Create-React-App) and a GRPC backend. We setup a reverse proxy (check out Envoy for example) so that the GRPC servicers are served from the same origin as the frontend, but with a certain prefix (In our example, / serves the frontend, and /grpc is the prefix for all GRPC servicers)

Then, when making your promise client you can specify a prefix in the URL.
var BillingService = new BillingPromiseClient('/grpc');

Once you set your cookies (either though some frontend request, or manually via Javascript), then they will be sent automatically on all requests to the same origin.

swuecho:
that is smart! thanks.

buehler:
For those who want to add the authorization header to each request, until interceptors are implemented, maybe you want to use my workaround:

(in typescript)
```typescript
import { HelloWorldServicePromiseClient } from '....';

export class GrpcClient {
[...]

Issue #176, opened 2018-04-19
Add ability to send cookies cross domain


aberasarte:
As far as I understand, there is no way to send cross domain cookies with the current grpc-web client implementation. Looking at the code, I see that grpc-web is using xhrio to send the requests to the server but I don't see any settable property nor option in the grpc client that would enable the xhrio "withCredentials" flag.

Would it be possible to add this feature? It would be really helpful for our current scenario.

stanley-cheung:
This feature is in the internal version of the library. We will be glad to add it here. How do you envision using this? Should this be a call-level option? Or should it be service/stub level?

wenbozhu:
Service/stub level. It's a no-op if calls are from the same origin.

aberasarte:
Thanks for the fast reply! Sevice/stub level works for me.

AaronTriplett:
Any updates on this? We're currently blocked from going to production due to this issue.

stanley-cheung:
Added an option in #604 to allow clients to enable `withCredentials` to `true`.

```
var client = new EchoServiceClient('localhost', null, {
  'withCredentials': true
});
```

But before I cut the next release including this fix, is it possible that someone can verify if this works for your use case? Coz I received a report that even with #604, cookie is not being attached to a cross-domain request. So I was wondering if we are still missing something.

hellraisercenobit:
I'm trying to do this with envoy but nothing seems to work like expected.
Someone have an envoy config sample to do this?
Cors allow_credentials is set server side.
withCredentials is set client side.
But Access-Control-Allow-Origin is always equal to '*' in the response headers.
So the request/response has been blocked by CORS policy.

bram-abe:
> Added an option in #604 to allow clients to enable `withCredentials` to `true`.
> 
> ```
> var client = new EchoServiceClient('localhost', null, {
>   'withCredentials': true
> });
> ```
> 
> But before I cut the next release including this fix, is it possible that someone can verify if this works for your use case? Coz I received a report that even with #604, cookie is not being attached to a cross-domain request. So I was wondering if we are still missing something.

Aha this solve my initial problem where the case is browser could read & receive Set-Cookie response header but can not store it. thx @stanley-cheung

vdhanan:
@stanley-cheung i've enabled `withCredentials` and the cookie is still not being attached to a cross-domain request. would you mind helping me resolve this?

```
    const unauthClientOpts = {
      unaryInterceptors: [versionInterceptor],
[...]

PR #498, opened 2019-03-14
feat: support for with-credentials xhr flag

n3squik:
Add support for xhr's with-credential flag ([specs](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials)).

A new optional parameter is added to clients constructors. If not set the default value is false, so it does not change the prior behavior.

Usage:
```
const WITH_CREDENTIALS = true;
const client = new SomeServiceClient('https://www.some-domain.com:8443', WITH_CREDENTIALS);
``` 

Related issues: #176, #351

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -647,7 +647,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
       printer->Print(vars, "options['format'] = '$format$';\n\n");
     }
     printer->Print(vars,
-                   "this.client_ = new grpcWeb.$mode$ClientBase(options);\n"
+                   "this.client_ = new grpcWeb.$mode$ClientBase(options, credentials);\n"
                    "this.hostname_ = hostname;\n"
                    "this.credentials_ = credentials;\n"
                    "this.options_ = options;\n");
@@ -1004,7 +1004,7 @@ void PrintServiceConstructor(Printer* printer,
       "  /**\n"
       "   * @private @const {!grpc.web.$mode$ClientBase} The client\n"
       "   */\n"
-      "  this.client_ = new grpc.web.$mode$ClientBase(options);\n\n"
+      "  this.client_ = new grpc.web.$mode$ClientBase(options, credentials);\n\n"
       "  /**\n"
       "   * @private @const {string} The hostname\n"
       "   */\n"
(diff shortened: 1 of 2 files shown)


jmzwcn:
seems this flag is unnecessary, give your credentials-value is enough directly.

n3squik:
> seems this flag is unnecessary, give your credentials-value is enough directly.

@jmzwcn could you elaborate a bit? where can I pass my credentials-value directly?

stanley-cheung:
This is necessary. The xhr.with-credentials flag has special meaning in the browser, which will include domain cookies to be attached to the request. We just never got around to adding an API for it.

jmzwcn:
Hop to avoid non-compatible and reduce too many parameters.

stanley-cheung:
@n3squik Thanks for your contributions. I might have to think about this a little more, especially around how exactly we are gonna do this. For the internal use case, we don't have the extra `credentials` parameter in the constructor. We tried to detect whether certain metadata is being set and that indicates that we also want to `.setWithCredentials(true)`. 

The challenge in here (the open-source use case) is that it's hard to define what that condition might be. We might have to do it with setting a specific parameter in the `options` parameter in the constructor. I really want to avoid adding another parameter the stub client constructor without an absolutely necessary reason. 

Let's keep this PR open - I will come back on this asap.

andrewtikhonov:
How long will this be open ? 

withCredentials: true is needed for browsers to trigger authentication as a result of XMLHttprequest. If not set, the browser is silent. Currently it's not possible to properly trigger any browser driven auth as a result of gRPC call, neither Negotiate (NTLM, Kerberos) nor Basic.

Something like below should be enough if you don't want to change the parameters.
```
    if (options.withCredentials) {
        xhrIo.withCredentials = true;
    }

```

Kupstas:
Any new information?

AaronTriplett:
Checking in on this, currently blocked. Any chance we can get this merged?

stanley-cheung:
Done in #604. Thanks for your contributions anyways. Sorry for the huge delay.

stanley-cheung:
Before cutting the next release, is it possible that someone can verify if this works for your use case?

```
var client = new EchoServiceClient('localhost', null, {
  'withCredentials': true
});
```

I received a report that even with #604, cookie is not being attached to a cross-domain request. So I was wondering if we are still missing something.

PR #501, opened 2019-03-20
Temporarily run 'bazel test' multiple times

stanley-cheung:
In some environment (could be a more recent version of Debian, but not entirely sure), the `bazel test` command being run from our master `kokoro.sh` will occasionally fail with these kind of error messages

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1                                                                   
Executing tests from //javascript/net/grpc/web:grpcwebstreamparser_test                                       
-----------------------------------------------------------------------------                                 
Auto configuration failed                                                                                     
140088820610880:error:2506406A:DSO support routines:DLFCN_BIND_FUNC:could not bind to the requested symbol name:dso_dlfcn.c:287:symname(OPENSSL_finish): /lib/x86_64-linux-gnu/libssl_conf.so: undefined symbol: OPENSSL_finish                                                                                                         
140088820610880:error:2506C06A:DSO support routines:DSO_bind_func:could not bind to the requested symbol name:dso_lib.c:294:                                                                                               
140088820610880:error:0E07606D:configuration file routines:MODULE_RUN:module initialization error:conf_mod.c:235:module=ssl_conf, value=ssl_sect, retcode=-1         
```

But the `bazel test` won't fail every time - it will run successfully 1 in maybe 8 times. Clearly the tests themselves are fine. Somehow the `bazel test` command is flaky, only in certain environment. 

I couldn't quite figure it out so I am temporarily making the `bazel test` command run 15 times and if it passes just once that's good enough for the purpose verifying a PR.

We should somehow figure this out later though.

### DIFF
--- a/scripts/kokoro.sh
+++ b/scripts/kokoro.sh
@@ -49,9 +49,26 @@ wget https://github.com/bazelbuild/bazel/releases/download/"${BAZEL_VERSION}"/ba
 chmod +x ./bazel-"${BAZEL_VERSION}"-installer-linux-x86_64.sh
 ./bazel-"${BAZEL_VERSION}"-installer-linux-x86_64.sh --user
 $HOME/bin/bazel version
-$HOME/bin/bazel test \
-  //javascript/net/grpc/web/... \
-  //net/grpc/gateway/examples/...
+
+# Run the bazel test command multiple times because of some flaky
+# behavior with libssl.so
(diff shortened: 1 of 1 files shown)


stanley-cheung:
@Yannic FYI

stanley-cheung:
#502 might fix this

PR #502, opened 2019-03-20
Attempt to fix flakiness of 'bazel test'

Yannic:
The cause of the failures might have been fixed in https://github.com/bazelbuild/rules_closure/pull/337.


scripts/kokoro.sh
```diff
@@ -48,28 +49,11 @@ BAZEL_VERSION=0.23.1
...
-    break
-  elif [ $i -eq $tries ]; then
-    exit 1
-  fi
-done
-set -e
-rm ./buildifier
 rm ./bazel-"${BAZEL_VERSION}"-installer-linux-x86_64.sh
+$HOME/bin/bazel version
+$HOME/bin/bazel test \
```
stanley-cheung: Can you add back the `$HOME/bin/bazel clean` line please? It makes running the script locally and repeatedly a little more robust. thanks.

Yannic: Sure. Done.

Yannic:
Yeah, Bazel is rather picky which parts of the environment are available in tests (If that's good or bad is left as an exercise to the reader).

Thanks for your review!

PR #512, opened 2019-03-22
Make client args `credentials` and `options` optional

jonahbron:
Currently the TypeScript generated clients require that `null` be passed to `credentials` and `options` if they're not being used.  It's easier to simply make those arguments optional.  I've left the `null` portion of the type though to maintain backwards compatibility.

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -639,8 +639,8 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
         "credentials_: null | { [index: string]: string; };\n"
         "options_: null | { [index: string]: string; };\n\n"
         "constructor (hostname: string,\n"
-        "             credentials: null | { [index: string]: string; },\n"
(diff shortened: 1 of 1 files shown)


jonahbron:
Not sure why Kokoro still hasn't completed its checks... but otherwise this PR is ready for review.

stanley-cheung:
Yea I have to apply the tag for the kokoro test to run, and the reason I haven't applied the tag is because I thought we did this and reverted a long while ago and I forgot the reason now and I haven't got a chance to dig out the history. I might remember wrong. Will get to this asap.

jonahbron:
Gotcha.  Thanks @stanley-cheung , no rush.

jonahbron:
@stanley-cheung I looked through the history of the master branch and found no evidence that those args were ever optional, outside of having been looser types in the past (i.e. `any`).

stanley-cheung:
After a bit of quick digging, what I had been thinking was #392 and #396. Looks like they are different than this. Just want to be careful on not trying to do this A, revert A, do A again and repeat :)

Will review this again soon.

PR #537, opened 2019-04-16
Fix grpc-web package typescript type error

tomykaira:
With grpc-web release 1.0.4, grpc-web package's index.d.ts casues
the following errors.

```
EchoServiceClientPb.ts:44:5 - error TS2345: Argument of type '(bytes: Uint8Array) => EchoResponse' is not assignable to parameter of type '(bytes: {}) => EchoResponse'.  Types of parameters 'bytes' and 'bytes' are incompatible.
    Type '{}' is missing the following properties from type
'Uint8Array': BYTES_PER_ELEMENT, buffer, byteLength, byteOffset, and 26
more.

44     EchoResponse.deserializeBinary
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is because `UInt8Array => T` argument is given for `{} => T` parameter.
Closure compiler type of `responseDeserializeFn` is `? => RESPONSE`,
which corresponds to `any => T` in the parameter context, not `{}`.

This fix changes `{}` to `any` in order to solve the above problem.

### DIFF
--- a/packages/grpc-web/index.d.ts
+++ b/packages/grpc-web/index.d.ts
@@ -6,7 +6,7 @@ declare module "grpc-web" {
     class MethodInfo<Request, Response> {
       constructor (responseType: new () => Response,
                    requestSerializeFn: (request: Request) => {},
-                   responseDeserializeFn: (bytes: {}) => Response);
+                   responseDeserializeFn: (bytes: any) => Response);
     }
   }
 


sergey-shambir:
It's better to use `Uint8Array` instead of `any` - both work for me.

tomykaira:
I intended to use the same type specification as that for google-closure-compiler https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/abstractclientbase.js#L48 .
Sorry, I do not know well about the library, and I am not sure whether it is possible to change `function(?): RESPONSE` to `function(Uint8Array): RESPONSE`.
If possible, I agree to change both `any` and `?` to `Uint8Array`.

sergey-shambir:
It was just a hint, I'm not a maintainer :)

stanley-cheung:
Thanks for the contributions. This will be superseded by #540.

PR #540, opened 2019-04-16
Method Derserializer should take Uint8Array

pnegahdar:
Currently running into issue:

```
Type '{}' is not assignable to type 'Uint8Array'.
```

This fixes it for me.


faraonc:
Is the fix released yet?

PR #542, opened 2019-04-23
fix: Add command to update package lists

duncandean:
Add `apt-get -qq update` to quietly update the debian stretch package list
so the `apt-get -qq install -y default-jdk` does not fail due to 404s.


stanley-cheung:
Can you rebase the PR please? There was a recent breakage.

duncandean:
Rebased 👍

PR #549, opened 2019-05-01
Fix TypeScript type issue

lorenz:
The generated code from 1.0.4 currently doesn't typecheck because this type is wrong.


lorenz:
It seems like there are two PRs already open for this fix (#537, #540), so no point in making another one.

Issue #554, opened 2019-05-08
Setting deadline to `Infinity` causes grpc-web to send "Infinitym" as grpc-timeout


CatEars:
I can set the deadline to `Infinity` in a server streaming call (maybe also unary?) and it will set the `grpc-timeout` header to `Infinitym`.

As far as I can tell it is cause by this line https://github.com/grpc/grpc-web/blob/d139a6ad994c9cc3ad619fdb1aa0a231a83e1d80/javascript/net/grpc/web/grpcwebclientbase.js#L220

stanley-cheung:
Ha, that's embarrasing :)

CatEars:
I wouldn't mind doing a pull request for this. Is the solution to just default to a certain value for `grpc-timeout` header when getting `Infinity` as input?

stanley-cheung:
I think 1. we should not accept a string as an input and 2. handle accepting *Infinity* (a number) passed as an argument

CatEars:
Solved with above PR

PR #556, opened 2019-05-10
Add fix for deadline of strings, NaN, Infinity and -Infinity

CatEars:
This will change `Infinity` to `Math.pow(2, 53) - 1` (`Number.MAX_SAFE_INTEGER` in ES6, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). Anything that becomes `NaN` will be discarded (for example setting deadline to a string or `NaN`), `-Infinity` will also be discarded.

I don't know if it is best to let bad values for deadlines slip past silently or not...

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientbase.js
+++ b/javascript/net/grpc/web/grpcwebclientbase.js
@@ -221,8 +221,7 @@ GrpcWebClientBase.prototype.processHeaders_ = function(xhr) {
       var maxSafeInteger = Math.pow(2, 53) - 1;
       timeout = maxSafeInteger;
     }
-    // isFinite discards NaN and -Infinity
-    if (isFinite(timeout) && timeout > 0) {
+    if (timeout > 0) {
       xhr.headers.set('grpc-timeout', timeout + 'm');
     }
   }
--- a/javascript/net/grpc/web/grpcwebclientbase.js
+++ b/javascript/net/grpc/web/grpcwebclientbase.js
@@ -217,8 +217,7 @@ GrpcWebClientBase.prototype.processHeaders_ = function(xhr) {
     var timeout = Math.round(deadline - currentTime);
     xhr.headers.remove('deadline');
     if (timeout === Infinity) {
-      // Polyfill for "Number.MAX_SAFE_INTEGER"
-      var maxSafeInteger = Math.pow(2, 53) - 1;
+      var maxSafeInteger = 2147483647; // 2^31 - 1
       timeout = maxSafeInteger;
     }
     if (timeout > 0) {
--- a/javascript/net/grpc/web/grpcwebclientbase.js
+++ b/javascript/net/grpc/web/grpcwebclientbase.js
@@ -216,10 +216,16 @@ GrpcWebClientBase.prototype.processHeaders_ = function(xhr) {
     var currentTime = (new Date()).getTime();
     var timeout = Math.round(deadline - currentTime);
     xhr.headers.remove('deadline');
-    if (timeout > 0) {
+    if (timeout === Infinity) {
+      // Polyfill for "Number.MAX_SAFE_INTEGER"
+      var maxSafeInteger = Math.pow(2, 53) - 1;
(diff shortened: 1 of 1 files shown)


CatEars:
In response to issue #554

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -216,10 +216,16 @@ GrpcWebClientBase.prototype.processHeaders_ = function(xhr) {
...
     var timeout = Math.round(deadline - currentTime);
     xhr.headers.remove('deadline');
-    if (timeout > 0) {
+    if (timeout === Infinity) {
+      // Polyfill for "Number.MAX_SAFE_INTEGER"
+      var maxSafeInteger = Math.pow(2, 53) - 1;
+      timeout = maxSafeInteger;
+    }
+    // isFinite discards NaN and -Infinity
+    if (isFinite(timeout) && timeout > 0) {
```
rogchap: Do we support IE 11 ? 
If `yes` we need to Polyfill `isFinite()`
If `no` we can use `isFinite()` and `Number.MAX_SAFE_INTEGER` (ie. no polyfill)
https://caniuse.com/#feat=es6-number

CatEars: Should i create a polyfill for isFinite, just in case? Not sure if IE 11 should be supported or not, but fixing the polyfill in this case should be as simple as:

```javascript
...
if (timeout !== NaN && timeout !== -Infinity && timeout > 0) { ... }
```

right?

Yannic: I feel like @rogchap is talking about `Number.isFinite`, while this is `window.isFinite` which was already part of ES3 (Section 15.1.2.5 of https://www-archive.mozilla.org/js/language/E262-3.pdf).
I don't think a polyfill is required.

If we add the polyfill, it's sufficient to check `timeout !== Infinity` and `timeout > 0`.
`timeout !== NaN` is always true, because, accoring to the IEEE standard, comparing any number to `NaN` is false (including `NaN`).
```javascript
if (timeout !== Infinity && timeout > 0) { ... }
```

CatEars: Thanks @Yannic for providing the info on `window.isFinite` vs. `Number.isFinite` and `NaN !== NaN`. I didn't know this so I think it was very helpful. Perhaps sticking with `isFinite()` and no polyfill is the way to go then?

CatEars: Or wait... Perhaps we don't need `isFinite()` at all then.

There is already a check for `Infinity` on line 219, `timeout > 0` removes `-Infinity` and as @Yannic says, comparing any number to `NaN` is always false so `timeout > 0` is false for `NaN`. Which I think means there are only positive numbers left, which should all be allowed. I believe that sounds like the way to go.

Yannic: You're right that `timeout > 0` is sufficient!. I must have missed the check for `Infinity` above.

That being said, I think we should set the timeout to `2147483647 // 2^31 - 1` to be on the save side. Servers might use 32-bit signed integers for the timeout which will overflow for `2^53 - 1` Also, pre-calculating constants and commenting them is, IMO, better than calculating them every single time they're needed (not that it makes a big difference in this case).

Otherwise LGTM.

CatEars: I added the above so now I think the code should be finished.

CatEars:
> Sorry for the delaying in reviewing this.

No problem, being an open source maintainer is tough work!
 
> Actually, in the gRPC protocol definition, if Timeout is omitted, it's assumed to be Infinite. So what we should be doing here is that, in L215, if "Infinity" is passed in, just don't set the "grpc-timeout" header. Does it make sense?

I think then that defaulting to `0` is a good idea? I pushed up such a change in 5650699.

PR #558, opened 2019-05-11
API Proposal - Client Interceptors

rogchap:


proposals/client-interceptors.md
```diff
@@ -0,0 +1,312 @@
...
+
+A `listener` object implements zero or more methods for intercepting inbound operations. The listener passed into a requester's `start` method implements all the inbound interception methods. Inbound operations will be passed through a listener at each step in the interceptor chain. Three usage patterns are supported for listeners:
+
+1. Pass the listener along without modification: `next(metadata, listener)`. In this case the interceptor declines to intercept any inbound operations.
+1. Create a new listener with one or more inbound interception methods and pass it to next. In this case the interceptor will fire on the inbound operations implemented in the new listener.
+1. Store the listener to make direct inbound calls on later. This effectively short-circuits the interceptor stack. Short-circuiting a request in this way will skip interceptors which have not yet fired, but will fire the listeners on any 'earlier' interceptors in the stack.
+
+*Do not modify the listener passed in. Either pass it along unmodified or call methods on it to short-circuit the interceptor stack.*
+
+The `listener` methods are:
```
rogchap: So, I'm not 100% convinced on the `listener` API for inbound operations. 
This has been ported from the NodeJS API; however I believe that this is over-complicated for the web implementation.
The way I see it, there only a few places where we need to intercept:

https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientbase.js#L66
```javascript
GrpcWebClientBase.prototype.rpcCall = function(
    method, request, metadata, methodInfo, callback) {
  // Intercept HERE onStart()
  // ...
  stream.on('data');   // Intercept HERE onData()
  stream.on('status'); // Intercept HERE onStatus()
  stream.on('error'); // Intercept HERE onError()
  // ...
  // Intercept HERE onSend()
  xhr.send(method, 'POST', payload);
  return stream;
};
```
The `serverStreaming` is similar.

This would simplify the `requester` object to:

```javascript
var requester = {
  onStart: function(metadata, next){},
  onSend: function(request, next){},
  onData: function(response, next){},
  onStatus: function(status, next){},
  onError: function(error, next){},
};
```

proposals/client-interceptors.md
```diff
@@ -0,0 +1,312 @@
...
+
+An interceptor is a function which takes an `options` object and a `nextCall` function and returns an `InterceptingCall` object:
+
+```javascript
+/**
+ * @param {Object} options
+ * @param {string} options.hostname The server host to call
+ * @param {Object} options.credentials Credentials used to authenticate with the server
+ * @param {MethodDescriptor} options.methodDiscriptor A container of properties describing the call (see below)
+ * @param {Function} nexCall Constructs the next interceptor in the chain
```
duncandean: nexCall -> nextCall 🙂

duncandean:
Thanks for this! It looks great. 😄

proposals/client-interceptors.md
```diff
@@ -0,0 +1,312 @@
...
+
+```javascript
+/**
+ * @param {string} name The name of the call, i.e. 'myCall' 
+ * @param {string} serviceName The name of the service, i.e. 'MyService'
+ * @param {string} path The full path of the call, i.e. '/MyService/MyCall'
+ * @param {MethodType} methodType One of two types: MethodType.UNARY or MethodType.SERVER_STREAMING
+ * @param {grpc.web.AbstractClientBase.MethodInfo} methodInfo Properties containing the serialize and deserialize functions
+ */
+MethodDescriptor(name, serviceName, path, methodType, methodInfo)
```
jatraman: wdyt about moving the existing class out of AbstractClientBase and creating grpc.web.MethodInfo that works for this and other use cases? 

Couple of reasons to consider this:
1. If grpc.web.MethodInfo also contains the path it should simplify the proto compiler as well (js_grpc_web_library).
2. It'll be confusing to have both MethodInfo and MethodDescriptor

proposals/client-interceptors.md
jatraman: A case can be made to keep the API design consistent with other platforms. That way developers don't need to re-learn concepts. For instance, I find it convenient that the Java API[1] is the same for Android and server. A ClientCall in Java roughly translates to InterceptingCall as defined in this proposal. So defining a Listener means the JS API looks similar to the Java API.

OTOH, the ObjC Interceptor API proposal does *not* define the concept of Listener and instead puts all methods on GRPCInterceptorInterface[2] instead.

[1]https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.html#start-io.grpc.ClientCall.Listener-io.grpc.Metadata-
[2]https://github.com/grpc/proposal/blob/9bcd06aae8e6f9122885711a602ecdc9af75c560/L50-objc-interceptor.md#interceptor-chain

proposals/client-interceptors.md
```diff
@@ -0,0 +1,312 @@
...
+```
+*Do not modify the `options.methodDescriptor` object. The MethodDescriptor passed to the interceptor options is not consumed by the underlying gRPC code and changes will only affect downstream interceptors*
+
+A `requester` object is a plain Javascript object implementing zero or more outbound interception methods:
+
+```javascript
+/**
+ * An interception method called before an outbound call has started.
+ * @param {grpc.web.Metadata} metadata The call's outbound metadata (request headers).
```
jatraman: Is grpc.web.Metadata a new class?

proposals/client-interceptors.md
```diff
@@ -0,0 +1,312 @@
...
+var constructorOptions = {
+    interceptorProviders: interceptorProviders,
+};
+var client = new MyServiceClient(hostname, credentials, constructorOptions);
+```
+The order of the InterceptorProviders will determine the order of the resulting interceptor stack.
+
+#### At call invocation
+
+If an array of interceptors is provided at call invocation via `options.interceptors`, the call will ignore all interceptors provided via the client constructor.
```
jatraman: Why ignore vS add to end of list? Do you have some concerns about interceptor ordering?

Interceptors like authorization, logging, metrics collection are likely to be defined and configured centrally. It would be unfortunate if a call site that just wants to add a custom interceptor needs to redefine this entire list again.

proposals/client-interceptors.md
rogchap: Yup, I think this would be a good idea. 
My initial thoughts on the implementation was to make minimal changes to the current API and CodeGen and provide backward compatibility (hence keeping `methodInfo` as-is). The reality may be that the refactor may not be compatible... we would need to make a call on backward compatibility support; if no then this would push this project to v2.0.0

proposals/client-interceptors.md
rogchap: No, I stole this from the TypeScript definition:
https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/index.d.ts#L3
I probably should have left it as `Object<string, string>`

proposals/client-interceptors.md
rogchap: I would expect that most interceptor authors to have there interceptor chains easily available to add at call invocation. 
Yes, the main reason is the ordering, or possibly you want a different `authorization` mechanism for that call etc.
I still think that this API has to override those at the client level, however maybe there is a second API that allows you to `add` interceptors to the chain? This would be useful for tools to inject interceptors; for example the grpc-web chrome DevTools.

proposals/client-interceptors.md
jatraman: fyi, https://github.com/grpc/grpc-web/pull/567 introduced grpc.web.MethodDescriptor and MethodType.

clehene:
Hey @stanley-cheung what's holding this PR? It would be useful to have a clean way to refresh client side tokens. Perhaps even provide a JWT interceptor with the distribution. Something like https://github.com/avast/grpc-java-jwt

kahirokunn:
I need grpc web client intercepter.
What’s holding this PR?

kahirokunn:
@duncandean how about this now ?

kahirokunn:
@duncandean hello...?

duncandean:
@kahirokunn I'm not a reviewer authorised with write access. This proposal probably also needs more thoughts and comments.

kahirokunn:
I see... 😢

oreno4649:
I think this PR is necessary

maguro:
How are things going on this?

stanley-cheung:
We will be picking this up soon, sorry for the delay.

stanley-cheung:
Hi @rogchap, and everyone else who is interested in interceptors:

First we are sorry it took us so long to get our act on this - resources, priorities, blablabla :) But we had not been idle on this - over the last many months, we have done some investigation internally to see how we can have a unified implementations of interceptors that will satisfy both our internal use case and the open source use case here. We have some designs and discussions internally to see how we will approach this. Also we identified some issues and requirements with the existing APIs that kind of get in our way for some of the features we plan internally.

So over the last few months we had done some background work to introduce a few concepts like [`grpc.web.Request`](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/request.js), [`grpc.web.MethodDescriptor`](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/methoddescriptor.js) that requires some delicate migration work, and a few other classes like `GenericClient` and `UnaryResponse`. Some of those come from the need to fulfill some requirements to integrate with some other internal frameworks. And at the same time we want to use those concepts for the Interceptors design too. 

We are at a stage where we are close to be able to put a PR forward in the coming days to show how we plan to implement interceptors here in this repo. I have started a thread over at #766 to talk about how the interceptors are designed and provided some examples there. This is still work-in-progress - we are still working through some issues. Please check these out and let us know how you think!

Thanks.

stanley-cheung:
Closing this. The new interceptors proposal have been detailed in #766 and have the code merged in #785. We will be working on a `1.1.0` release soon.

PR #567, opened 2019-05-24
Introducing MethodDescriptor

stanley-cheung:
Related to #558. We are experimenting with introducing the `MethodDescriptor` class to eventually replace `AbstractClientBase.MethodInfo`. This will match the Client Interceptor proposal and will eventually be more flexible and expressive than the current MethodInfo definition. 

Creating this to start running some tests


stanley-cheung:
Merging this as the code is submitted internally.

PR #577, opened 2019-06-20
Add `COPY` operation in client Dockerfile for local test

zhouqiangWang:
Add `COPY files` from host into container in `commonjs_client/Dockerfile` and `ts_client/Dockerfile`.
Make local test easier.


stanley-cheung:
Thanks for the PR. I understand the intention. But if someone intends to use this as a baseline and constantly be changing the code and re-building the docker image, I feel that it would be more appropriate to have your own copy of this Dockerfile and go from there. Sorry won't be merging this.

PR #595, opened 2019-07-18
Updating for new Envoy versions

mitar:
Since 1.10.0 `enabled` is deprecated. Removing it works because the default is for filter to be enabled.

More information:
 * https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated
 * https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-field-route-corspolicy-enabled
 * https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-field-route-corspolicy-filter-enabled


gary-lo:
Same issue as: https://github.com/grpc/grpc-web/pull/593
Different example.

Issue #596, opened 2019-07-20
MethodDescriptor Introduced in #567 is broken


clehene:
/cc @stanley-cheung 
I'm running with ` --grpc-web_out=import_style=commonjs+dts,mode=grpcweb`

Here's a generated MethodDescriptor: note `optin_pb.OptInRequest` instead of `proto.x.protobuf.v1alpha2.OptInRequest`

```javascript
/**
 * @const
 * @type {!grpc.web.MethodDescriptor<
 *   !proto.x.protobuf.v1alpha2.OptInRequest,
 *   !proto.google.protobuf.Empty>}
 */
const methodDescriptor_OptInService_OptIn = new grpc.web.MethodDescriptor(
  '/x.protobuf.v1alpha2.OptInService/OptIn',
  grpc.web.MethodType.UNARY,
  optin_pb.OptInRequest,
  google_protobuf_empty_pb.Empty,
  /** @param {!proto.x.protobuf.v1alpha2.OptInRequest} request */
  function(request) {
    return request.serializeBinary();
  },
  google_protobuf_empty_pb.Empty.deserializeBinary
);
```

I'm unable to compile grpc-web anymore (no build instructions either..), however the problem seems around these lines

https://github.com/grpc/grpc-web/pull/567/files#diff-2e729f9088e537bfeb6fefc318963a96R1078
https://github.com/grpc/grpc-web/pull/567/files#diff-2e729f9088e537bfeb6fefc318963a96R1389

This line  could use a comment to explain what's the goal of `method->output_type()->file() != file`

```javascript
if (import_style == ImportStyle::COMMONJS &&
            method->output_type()->file() != file) {
```

buehler:
Yep, I'm also experiencing this error. It seemed to work in 1.0.4

travikk:
Same here, used to work with 1.0.4

stanley-cheung:
Sorry for being late to look into this.

I just looked into this - but it appears that the issue may run deeper than I thought. At first I thought it's some DOA issue that borking everything but it's not.

I run the 1.0.5 plugin on the helloworld example and it works. Then I realized the helloworld example uses `--grpc-web_out=import_style=commonjs`.

So I tried the Echo ts-example next, since that uses `--grpc-web_out=import_style=commonjs+dts`. But it works too. And I verified that the generated code (from the 1.0.5 plugin) `echo_grpc_web_pb.js` has the new `MethodDescriptor` stuff, like

```
[...]

PR #599, opened 2019-07-30
fix wrong package name of input type

lqs:
When the input type of a method is defined in another file, the code generator generates a wrong package name for that type. This patch fixes this issue by checking both input and output types.


stanley-cheung:
This fixes #596

Issue #574, opened 2019-06-04
getStreamingResponseHeader errors when header not in Access-Control-Expose-Headers


pnegahdar:
https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientreadablestream.js#L122

The invocations fail  when the grpc-status is exposed in the payload (not header) per the spec. Since the `grpc-status` and `grpc-message` headers are no longer declared in the `Access-Control-Expose-Headers`  header, accessing it from the Xhrio lib will give an error instead of returning `null`

These are related to the issues pointed out in #518  & https://github.com/improbable-eng/grpc-web/issues/396

stanley-cheung:
@pnegahdar Sorry for being late to look into this.

I think what you said make sense, but how do I get into a situation where "the grpc-status is exposed in the payload (not header) per the spec"?

Vuhag123:
Are you still facing this issue?

Issue #544, opened 2019-04-23
Aborting ClientReadableStream results in error event being fired


neilcarpenter:
I believe this is handled here: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientreadablestream.js#L199

Calling `.abort()` results in `"error"` event being triggered, this is problematic in situations such as this:
```
// User interacts with UI, we trigger grpc request:
const stream = myService.method();

// We setup error handler in request, displaying error in UI to user
stream.on("error", err => {
  displayErrorMessageToUser();
});

// User clicks "cancel" button to abort grpc service request
stream.cancel();
```
In this example, the user will be shown an "error" message after they have aborted a request.

This issue is compounded by the fact that `err.code` passed to the `"error"` callback uses `grpcWeb.StatusCode.UNAVAILABLE` instead of `grpcWeb.StatusCode.ABORTED` meaning we can't even reliably check in the error callback the nature of the error being fired.

In my opinion a client-initiated cancellation of a gRPC-web request should _not_ trigger an error.

stanley-cheung:
hm, that makes sense.

PR #608, opened 2019-08-03
Fix status and error callbacks

stanley-cheung:
Changes introduced this PR:

 - If user initiated a `cancel()` on a stream, don't trigger an `error` callback for the `ABORTED` event. This is for #544 
 - Make sure we check whether the header key `grpc-status` and `grpc-header` are present in `getResponseHeaders()` before trying to access them. This is for #574 
 - Update the Node server example to use the `@grpc/grpc-js` package rather than the `grpc` package.
 - Add `cancel` to be a command in the Echo example. This will call `stream.cancel()`.


leoluk:
@stanley-cheung Could you please do a release which includes this fix?

PR #611, opened 2019-08-13
Change plugin name to grpc_web in README

gsalisi:
`protoc` seems to be looking for the grpc_web plugin name instead of the grpc-web.

Current readme instructions results into:
```
protoc-gen-grpc_web: program not found or is not executable
````

Tested with protoc@3.9.1 on OSX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ For example, in MacOS, you can do:
 
 ```
 $ sudo mv ~/Downloads/protoc-gen-grpc-web-1.0.6-darwin-x86_64 \
-  /usr/local/bin/protoc-gen-grpc-web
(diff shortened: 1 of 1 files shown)


stanley-cheung:
I believe how it works is, if your plugin is named `protoc-gen-<whatever>`, then you use it with `protoc -I=... ... --<whatever>_out=<params>`

So in the case where the plugin is named `protoc-gen-grpc-web`, you use it with `protoc -I=... ... --grpc-web_out=<params>`.

There is no requirement that the `protoc` command needs to be `--grpc_web_out=<params>`

gsalisi:
Ah, thanks for pointing that out. I couldn't find where that piece is from. :)

Issue #569, opened 2019-05-27
Error code 14 on 401 unauthorized


dasois:
The grpc-web client transforms a 401 into status code 14 .

The [definition suggests](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), that it should be error code 16 in this case.

I have the requirement to handle 401 errors differently than if the backend would be unavailable per se.

stanley-cheung:
Sorry for the late reply. But where did you see that 401 got turned into code 14?

Here: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/statuscode.js#L156 we mapped 401 to 16.

harmangakhal:
I have a similar problem; I've set up a web application using 
grpc-web/envoy and have an [external authorization ](https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/ext_authz_filter)
HTTP filter than performs JWT authentication of the client requests before allowing them through to the grpc server. When it returns,
a 401 response on JWT authentication failure, I think the response is not propagated because https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientreadablestream.js#L198-L207 does not handle [http_error](https://developers.google.com/closure/library/docs/xhrio).
It just default's to `UNAVAILABLE` so I get a
`Object { code: 14, message: "Http response at 400 or 500 level" }` response but
the HTTP POST response status code header returns a 401.
Seems like it's getting caught in, https://github.com/grpc/grpc-web/blob/e2f999a0df483336660406e970b1672ab1bc2a3e/javascript/net/grpc/web/grpcwebclientreadablestream.js#L212-L218
because the error message comes from,
https://github.com/google/closure-library/blob/4f1e354899494227cdf6ddeac9fa1f5b931b39fb/closure/goog/net/errorcode.js#L115-L116.

Is this intended behavior? Must my authentication server be a grpc server?

asv:
Have a similar problem as @haberman. @stanley-cheung could you help us? This is a serious production blocker for my team :(

asv:
@haberman I found a fast & dirty workaround.
From my ext_authz handler:
```golang
func (s *server) Check(ctx context.Context, req *envoyauthz.CheckRequest) (*envoyauthz.CheckResponse, error) {
	headers := req.GetAttributes().GetRequest().GetHttp().GetHeaders()
	sessionCookie, err := parseCookie(headers["cookie"], sessionCookieName)
	if err != nil {
[...]

PR #619, opened 2019-08-23
Return specific grpc status code on http error

harmangakhal:
Resolves the issue I mentioned in #569.


echoma:
Thanks a lot.  really need this fix, user authentication in Istio cant work properly without this fix.

@stanley-cheung When will the grpc-web team make a new release? cant wait :)

PR #625, opened 2019-09-03
uppercase only the filename, not the directory

asuffield:
Currently this translates "foo/bar.proto" into
"Foo/barServiceClientPb.ts". After this change it will be the correct
"foo/BarServiceClientPb.ts" instead.


stanley-cheung:
I believe this is similar to #621 - please check.

asuffield:
Yes, that also fixes this bug.

PR #655, opened 2019-10-17
Use closure compiler from npm in build.js

vbfox:
This use the script npm generate in `.bin` inside `build.js` so that the version align and a global install isn't necessary.

### DIFF
--- a/packages/grpc-web/scripts/build.js
+++ b/packages/grpc-web/scripts/build.js
@@ -42,7 +42,8 @@ const closureArgs = [].concat(
   ]
 );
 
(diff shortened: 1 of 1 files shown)


vbfox:
If the current directory contains the `node_modules` folder (or is a subdirectory) there is no difference, `npx` will essentially find the same file specified here in `.bin` and run it. 

Accessing it directly only changes the behavior if the current directory is somewhere else (`npx` will get the latest version from npm)

In the end it's you choice there is not much difference (A little bit more security on what version is resolved that's all)

Issue #543, opened 2019-04-23
`end` event occasionally STILL not emitted?


sirudog:
Hi,

I have upgraded my project to 1.0.4 so that I can eliminate the workaround taken from issue https://github.com/grpc/grpc-web/issues/289 needed for being notified about end of server-side streaming. In my project I use nginx (grpc-web enabled), C# gRPC service and a common.js+dts client.

**Unfortunately, this bug still seems to be not fixed: sometimes the `end` event is just not fired.**

Before raising this issue, I installed the echo example (https://github.com/grpc/grpc-web/tree/1.0.4#advanced-demo-browser-echo-app) using win10 + docker desktop.  I started it in chrome, but I never got a console message of `stream end signal received`, so the issue there is even worse. Only `Received metadata` message is logged on `status` event, as the screenshot below shows.
![image](https://user-images.githubusercontent.com/1550561/56584799-d313a000-65dc-11e9-8296-969c60c4f6cc.png)

I believe the end event handler in below code sample taken from the echo example should have also been called at end of each streaming response.
```
o.EchoApp.prototype.repeatEcho = function(e, t) {
        o.EchoApp.addLeftMessage(e),
        t > o.EchoApp.MAX_STREAM_MESSAGES && (t = o.EchoApp.MAX_STREAM_MESSAGES);
        var r = new this.ctors.ServerStreamingEchoRequest;
        r.setMessage(e),
        r.setMessageCount(t),
        r.setMessageInterval(o.EchoApp.INTERVAL);
        var n = this.echoService.serverStreamingEcho(r, {
            "custom-header-1": "value1"
        })
          , g = this;
        n.on("data", function(e) {
            o.EchoApp.addRightMessage(e.getMessage())
        }),
        n.on("status", function(e) {
            g.handlers.checkGrpcStatusCode(e),
            e.metadata && (console.log("Received metadata"),
            console.log(e.metadata))
        }),
        n.on("error", function(e) {
            o.EchoApp.addRightMessage("Error code: " + e.code + ' "' + e.message + '"')
        }),
        n.on("end", function() {
            console.log("stream end signal received")
        })
    }
```
I would appreciate if someone could check this.

sirudog:
Anyone has the same issue with the echo example? Or with 'end' event of server-side streaming in general?

martinblostein:
I have noticed the same issue with both the echo example, and originally in my own code.

annismckenzie:
[...]

Issue #361, opened 2018-11-08
ERR_INCOMPLETE_CHUNKED_ENCODING 200 (OK) on idle stream


jonno317:
I have a few streams running through an envoy proxy and am seeing the same behavior mentioned in #309 .

Specifically after 5 minutes streams timeout with a POST net::ERR_INCOMPLETE_CHUNKED_ENCODING 200 (OK) error. I do have the max_grpc_timeout set in the envoy config file in the way mentioned in the documentation from #309. 

I don't see any other known bugs around server side streaming. Are there potentially other settings or is maybe there's a gotcha of an idle timeout of 5 minutes even with the grpc timeout turned off (I'm fairly certain the streams are only dying after 5 minutes of inactivity)?

SebDE:
I ran into a similar problem. I would like to use the server-stream for a listener-pattern. In detail I would like to register the client via a server-stream to a listener on a grpc-server and get continuous updates. But after some time of inactivity the stream died like @jonno317 described. Are there any options for long running streams, or better for infinitely running server-streams?

jonno317:
@SebDE I did come across a setting on the envoy side of things that seems to be mitigating it. I haven't thoroughly tested yet, but I also haven't seen the issue again since I added it.

I believe the config from the example can be modified as follows to support long-lived streams.

```yaml
listeners:
  - name: listener_0
    address:
      socket_address: { address: 0.0.0.0, port_value: 8080 }
    filter_chains:
    - filters:
      - name: envoy.http_connection_manager
        config:
          codec_type: auto
          stat_prefix: ingress_http
          stream_idle_timeout: 0s  #<-- ADD THIS
          route_config:
            name: local_route
            virtual_hosts:
            - name: local_service
              domains: ["*"]
              routes:
              - match: { prefix: "/" }
                route:
                  cluster: echo_service
                  max_grpc_timeout: 0s
          http_filters:
          - name: envoy.grpc_web
          - name: envoy.router
  clusters:
  - name: echo_service
    connect_timeout: 0.25s
    type: logical_dns
    http2_protocol_options: {}
    lb_policy: round_robin
    hosts: [{ socket_address: { address: node-server, port_value: 9090 }}]
```

Hopefully someone else can chime in on whether there's a downside to that combination of settings.

weilip1803:
[

chiamtc:
[...]

PR #657, opened 2019-10-17
Ensure that the end callback is called

vbfox:
This should fix the case where a server stream was closed by some element on the network (Like envoy proxy with a timeout configured) and grpc-web didn't report any even (Neither 'error' or 'end') this fixes for us symptoms similar to #543 (We also see ERR_INCOMPLETE_CHUNKED_ENCODING more rarely like #361 but don't have a good solution yet) 

Previously the end callback was called by checking the state at the end of the `READY_STATE_CHANGE` event.

The problem with this approach is that there are multiple early exit cases in this function and when they happen the consumer is never informed of the end of the stream.

The new approach is to only dispatch the event when the `COMPLETE` event is received (Except if an error is dispatched) so the last event is always either `'error'` or `'end'`).

I'd like feedback on this last point as I did the 'only error or end' part by feeling but couldn't find any specification of how it's supposed to work, the other choices being to send error + end when an error is detected.

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientreadablestream.js
+++ b/javascript/net/grpc/web/grpcwebclientreadablestream.js
@@ -37,7 +37,6 @@ const EventType = goog.require('goog.net.EventType');
 const GrpcWebStreamParser = goog.require('grpc.web.GrpcWebStreamParser');
 const StatusCode = goog.require('grpc.web.StatusCode');
 const XhrIo = goog.require('goog.net.XhrIo');
-const XmlHttp = goog.require('goog.net.XmlHttp');
 const events = goog.require('goog.events');
 const googCrypt = goog.require('goog.crypt.base64');
 const googString = goog.require('goog.string');
--- a/javascript/net/grpc/web/grpcwebclientreadablestream.js
+++ b/javascript/net/grpc/web/grpcwebclientreadablestream.js
@@ -179,14 +179,6 @@ const GrpcWebClientReadableStream = function(genericTransportInterface) {
         }
       }
     }
-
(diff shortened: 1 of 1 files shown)


stanley-cheung:
Looks like the require `XmlHttp` is no longer used anywhere else in the file, please remove it. thanks.

```
ERROR: /tmpfs/src/github/grpc-web/javascript/net/grpc/web/BUILD.bazel:142:1: Compiling 127 JavaScript files to javascript/net/grpc/web/grpcwebclientbase_test_bin.js failed (Exit 1)
javascript/net/grpc/web/grpcwebclientreadablestream.js:40: ERROR - extra require: 'XmlHttp' is never referenced in this file
const XmlHttp = goog.require('goog.net.XmlHttp');
                ^
  ProTip: "JSC_EXTRA_REQUIRE_WARNING" or "extraRequire" or "legacyGoogScopeRequire" can be added to the `suppress` attribute of:
  //javascript/net/grpc/web:grpcwebclientreadablestream
  Alternatively /** @suppress {legacyGoogScopeRequire} */ can be added to the source file.

1 error(s), 0 warning(s), 97.3% typed

INFO: Elapsed time: 95.875s, Critical Path: 65.77s
INFO: 537 processes: 396 linux-sandbox, 141 worker.
FAILED: Build did NOT complete successfully
//javascript/net/grpc/web:grpcwebclientbase_test                      NO STATUS
//javascript/net/grpc/web:grpcwebstreamparser_test                       PASSED in 1.0s
```

As to your last question, Yes I believe the early exits is a problem we need to fix. But on the other hand, I am not very sure about when exactly the underlying `end` event fires either.

PR #665, opened 2019-10-24
Add config for Bazel CI

Yannic:
This will ensure gRPC-Web is compatible with the latest Bazel.
If we enable presubmit testing, we can remove Bazel from Kokoro
and speed-up CI.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -15,6 +15,6 @@ platforms:
 
   ubuntu1804:
     build_targets:
-      - "//..."
+      - //...
     test_targets:
       - //...
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,0 +1,20 @@
+---
+# TODO(yannic): Enable buildifier and test on Windows and RBE (both unsupported by rules_closure).
+platforms:
+  macos:
+    build_targets:
+      - //...
+    test_targets:
+      - //...
+
+  ubuntu1604:
+    build_targets:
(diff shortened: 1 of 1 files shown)


stanley-cheung:
@Yannic Thanks for contributing to anything Bazel related :) 

So what will this do? Next time when a PR is created, it will go out and find all the bazel targets under this repo and run `bazel build` and `bazel run`?

Yannic:
Once this is merged and the Bazel CI folks created the pipeline, this will build and test all Bazel targets 1. once a day against the latest Bazel release, and 2. periodically test against Bazel@HEAD to check for compatibility with upcoming breaking changes (e.g. https://github.com/bazelbuild/bazel/issues/8922, which I'll send a PR for soon-ish).
It's also possible to run Bazel CI for every gRPC-Web PR, but that requires to add a webhook to Github so the CI knows about the PR :). You can comment on https://github.com/bazelbuild/continuous-integration/issues/858 and they'll walk you through that process.

.bazelci/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+
+  ubuntu1604:
+    build_targets:
+      - //...
+    test_targets:
+      - //...
+
+  ubuntu1804:
+    build_targets:
+      - "//..."
```
stanley-cheung: Slight inconsistency on whether we need double quotes?

Yannic: I shall not copy&paste from more than one source...

Issue #399, opened 2018-11-30
Decoding grpc-status-details-bin ?


jscti:
Hi there

Server side, we send an error withDetails (in GO) : 
```
st2 := status.New(codes.Aborted, "failed to map")
    st2.WithDetails(&Error{
        Code:     1,
        Message: "error",
    })
```

Client side (js / typescript), we need to retrieve these details somehow but all we receive is a serialized string (`trailers.get('grpc-status-details-bin')` ) .. how can we decode it ?

Thanks

johanbrandhorst:
Go encodes the metadata as base64 after marshalling the proto message, so you should be able to base64 decode followed by a unmarshalling of the data into a `google.rpc.Status` type.

johanbrandhorst:
See https://github.com/grpc/grpc-go/blob/ca62c6b92c334f52c7e48b7cbf624f3b877fb092/internal/transport/http_util.go#L317 for how its done in Go clients. Should be very applicable to JS clients too. We may need to add some convenience function for this, but for now this is what you will need.

jscti:
Thanks for the (very) quick reply.

DIdn't recognize a base64 string, but yeah, decoded it this way and it's OK ;)

Thanks

jscti:
Struggling for a few hours now, I can't figure out how to parse the base64decoded string.
Nothing seems related on node_modules/grpc or node_modules/grpc_web_client. Can't find either any grpc/status look-alike file.

If you have any clue ^^

Server side : 
```
st2 := status.New(st, "test")
    st2, _ = st2.WithDetails(&gateway_public.Error{
        ErrorType: gateway_public.ErrorType_SPONSORSHIPCODE_CAPTCHA_VERIFICATION_FAILED,
        Message:   "A",
    }, &gateway_public.Error{
        ErrorType: gateway_public.ErrorType_SPONSORSHIPCODE_CAPTCHA_VERIFICATION_FAILED,
        Message:   "B",
    }, &gateway_public.Error{
        ErrorType: gateway_public.ErrorType_SPONSORSHIPCODE_CAPTCHA_VERIFICATION_FAILED,
        Message:   "C",
    })
```

Client side, I receive :
```
test1
(type.googleapis.com/gateway.public.ErrorA1
(type.googleapis.com/gateway.public.ErrorB1
(type.googleapis.com/gateway.public.ErrorC
```

johanbrandhorst:
[...]

PR #667, opened 2019-10-28
Proposal: Add metadata in gRPC error

at-ishikawa:
# What
* Add trailers as metadata or response headers as metadata in Error

# What is not 
* metadata for response is not included

# Why
Main motivation is to get grpc-status-details-bin on clients, discussed in #399 .
* My 1st thought was decode grpc-status-details-bin and include details in errors. But
    * This metadata is not defined in [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md). So this is not official and could be different/updated in implementations in each languages/libraries.
    * Even though details field of status is included in [status.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto#L93), grpc-status-details-bin returned in grpc-go is not encoded details, but encoded [status](https://github.com/grpc/grpc-go/blob/eca11cb9e47ddbe78dc0819ce345cb283450a135/internal/transport/handler_server.go#L210).
* If status event is subscribed, metadata from trailers can be retrieved. But there is no way using PromiseClient.
* By this implementations, custom metadata defined in gRPC spec can also retrieved.


at-ishikawa:
kokoro hasn't finished running, though I created about a day ago. Can someone check this?

stanley-cheung:
Yes I will get to this as soon as possible.

PR #670, opened 2019-10-29
Revert "Proposal: Add metadata in gRPC error"

stanley-cheung:
Reverts grpc/grpc-web#667

Sorry didn't catch the issue: `Metadata` wasn't properly defined in `error.js`.

### DIFF
--- a/javascript/net/grpc/web/error.js
+++ b/javascript/net/grpc/web/error.js
@@ -37,7 +37,4 @@ Error.prototype.code;
 /** @export {(string|undefined)} */
 Error.prototype.message;
 
-/** @export {{Metadata}} */
-Error.prototype.metadata;
-
 exports = Error;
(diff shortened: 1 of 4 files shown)


stanley-cheung:
@at-ishikawa FYI. I am getting some errors like

```
ERROR - [JSC_TYPE_MISMATCH] actual parameter 1 of callback does not match formal parameter
found   : {code: ?, message: ?}
required: (module$contents$grpc$web$Error_Error|null)
      callback({
               ^

1 error(s), 0 warning(s), 94.5% typed
```

stanley-cheung:
It appears that we need to add
```
const Metadata = goog.require('grpc.web.Metadata');
```
to `error.js`. And need to fix the JSDoc line to

```
/** @export {(?Metadata|undefined)} */
```

PR #671, opened 2019-10-29
Add metadata to error callback

stanley-cheung:
Roll forward of #667 after fixing JSDoc


stanley-cheung:
@at-ishikawa FYI

at-ishikawa:
@stanley-cheung Sorry for missing this and thank you for fixing the issue. 
The errors you shared in #670 are those when we run on browser, correct?
I think the reason why I didn't get is, I'd changed the compilation level of google closure compiler to SIMPLE and debugged it, but I'll try to reproduce it.

Issue #632, opened 2019-09-10
Unary callback executed multiple times on gRPC error


ascherkus:
I notice on HTTP/1.1 200 responses with a valid gRPC error that the callback is being executed three times. I believe it's a result of having a non-OK status code, which leads to executing the callback twice on "status" events and once more on the "error" event in `grpcwebclientbase.js`.

I found this to be surprising behavior (I expected at most a single non-null error callback), and since the callback is a required parameter I either have to use a local variable to track whether a non-null error has been previously encountered to prevent triggering error logic multiple times _or_ pass in no-op callback and use the returned `ClientReadableStream` object.

Example server code (Go):
```
func (s *server) Foo(context.Context, *foo.FooRequest) (*foo.FooResponse, error) {
	return nil, status.Error(codes.Unimplemented, "unimplemented")
}
```

Example client code (Typescript):
```
let client = new FooClient("http://localhost:8888");
let req = new FooRequest();
client.foo(req, null, function(err, resp) {
  console.log(err);
  console.log(resp);
  console.trace();
});
```

Console output:
```
{code: 12, message: "unimplemented"}
null
[ stack trace where "status" callback is triggered ]
{code: 12, message: "unimplemented"}
null
[ stack trace where "status" callback is triggered ]
{code: 12, message: "unimplemented"}
null
[ stack trace where "error" callback is triggered ]
```

I'm curious whether this is intended behavior or not (the double-status event in particular seems odd).

If not, I've got a few ideas:
1. Make the callback an optional parameter
2. Remove the callback entirely to support/force a consistent coding approach (i.e., same as streaming)
3. Don't execute the callback for status updates (i.e., only execute on "data" or "error")
4. Call the callback at most once on the first non-OK status/error response (whichever comes first)

Environment details:
  - protoc-gen-grpc-web-1.0.6-darwin-x86_64 
  - protoc 3.7.1
  - `--js_out=import_style=commonjs:$OUT`
  - `--grpc-web_out=import_style=typescript,mode=grpcweb:$OUT`
  - "google-protobuf": "^3.10.0-rc.1"
  - "grpc-web": "^1.0.6"
  - Server using `github.com/improbable-eng/grpc-web/go/grpcweb`

stanley-cheung:
This is probably fixed in #608, that hasn't been released yet. Will be included in the next release, namely `1.0.7`.

vsile:
On 01 May 2020 I steel have the issue.

[...]

PR #695, opened 2019-12-19
Fix issue 632 (double execution of callback)

hfinger:
This fixes that a grpc-status in the response header resulted in double execution of the callback that was provided to the rpcCall (https://github.com/grpc/grpc-web/issues/632)

In the echoAbort exampl to test grpc error response codes, this bug (https://github.com/grpc/grpc-web/issues/632) was not hit, because the used server library (grpc-js) is not fully compliant with other grpc server libraries. It seems that the grpc-js library is not sending grpc-status response headers in the same way as the golang, c++ implementations or that these headers are removed by envoy if grpc-js is used (this needs to be confirmed separately). In my tests, I found that simply changing the npm package to the native "grpc" library in the provided examples would result in the bug mentioned above. Therefore, I suggest to switch to the grpc npm package in the examples (i.e. grpc-js is still in beta).

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientreadablestream.js
+++ b/javascript/net/grpc/web/grpcwebclientreadablestream.js
@@ -229,7 +229,7 @@ const GrpcWebClientReadableStream = function(genericTransportInterface) {
         });
         errorEmitted = true;
       }
-      if (self.onStatusCallback_) {
+      if (!errorEmitted && self.onStatusCallback_) {
         self.onStatusCallback_(/** @type {!Status} */ ({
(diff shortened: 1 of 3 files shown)


hfinger:
Please could some maintainer let me know how I could proceed to get this pull request merged? Thanks.

stanley-cheung:
@hfinger Please follow the instructions above to signed the CLA. thanks!

hfinger:
I already signed the CLA on 19th December 2019. Not sure how I can proceed now. If I click now again on the EasyCLA link and click on "open CLA" then I am redirected to this page.

PR #738, opened 2020-02-25
Avoid double slash in url when client hostname has tailing slash

hanabi1224:
When GRPC client hostname has tailing '/', the final request url contains '//' which cannot be processed properly by some servers,  e.g.  grpc-dotnet-web


stanley-cheung:
I am so sorry. I keep looking at this and thought, how would this compile? I will have to look up what this `URL` class is and how it will compile, in C++. Only did I suddenly realize - it's _generated_ code. This is a _javascript_ class. *facepalm*.

Thanks for the contribution. But tests failed. I believe there's a missing comma after `toString()`.

Sorry about the delay.

hanabi1224:
Good catch! Fixed, please take another look

Fl0pZz:
@stanley-cheung @hanabi1224  this PR drop support uri like: `https://example.com/api/grpc/v1/...`

See examples:
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

stanley-cheung:
@Fl0pZz sorry I am a bit unclear on this - do you mean that you are passing the hostname parameter with the 3 dots at the end?

Fl0pZz:
> @Fl0pZz sorry I am a bit unclear on this - do you mean that you are passing the hostname parameter with the 3 dots at the end?

Nope :) 
I mean host name like this:
Hostname: `https://example.com/api/grpc/v1`

And in browser dev tools i see request: `https://example.com/api/grpc/v1/MyProject.listEntities`

stanley-cheung:
@Fl0pZz and what's wrong with the above?

Fl0pZz:
> @Fl0pZz and what's wrong with the above?

From mdn: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

```js
const hostname = 'https://example.com/api/grpc/v1';
const uri = new URL('/MyProject.listEntities', hostname);

console.log(uri.toString()) // -> https://example.com/MyProject.listEntities
```

however i expected `https://example.com/api/grpc/v1/MyProject.listEntities`

Issue #447, opened 2019-01-30
Failed to compile. 'proto' is not defined (also 'COMPILED')


argoyb:
I'm using create-react-app with typescript and grpc-web, grpc & proto generated with command:
```
$(protoc) -I $(proto_dir) $(proto_dir)/ride.proto --js_out=import_style=commonjs:./rider/typescript/src/proto --grpc-web_out=import_style=commonjs+dts,mode=grpcwebtext:./rider/typescript/src/proto
```

App is not able to start, it throws tons of errors of form
```
Failed to compile.

./src/proto/ride_pb.js
  Line 37:     'proto' is not defined     no-undef
  Line 40:    'proto' is not defined     no-undef
  Line 41:    'COMPILED' is not defined  no-undef
  Line 42:     'proto' is not defined     no-undef
 ...
```

argoyb:
I've tried to add `src/proto` to `exclude` section of `package.json` to exclude generated code from compilation but that didn't work. Probably because of following [thing](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html):
```
Any files that are referenced by files included via the "files" or "include" properties are also included. Similarly, if a file B.ts is referenced by another file A.ts, then B.ts cannot be excluded unless the referencing file A.ts is also specified in the "exclude" list.
```

argoyb:
Was able to bypass this problem by adding `/* eslint-disable */` to the top of generated `./src/proto/ride_pb.js`. However it resulted in another error:
```
Failed to compile.

<path>/rider/typescript/src/proto/ride_grpc_web_pb.js
Type error: Cannot compile namespaces when the '--isolatedModules' flag is provided.  TS1208

    10 |
    11 |
  > 12 | const grpc = {};
       | ^
    13 | grpc.web = require('grpc-web');
    14 |
    15 | const proto = {};

```

argoyb:
And I was able to bypass `Cannot compile problem` by adding `// @ts-ignore` to line 11 (above `const grpc = {};`)

stanley-cheung:
Sorry for the late reply. TypeScript support is 'experimental' right now. We would definitely welcome contributions. Seems like you have resolved your issue?

argoyb:
Well, I wouldn't call it "resolved" - I basically disabled typescript compiler check and eslint check. But true, it made project compile

caseylucas:
[...]

PR #752, opened 2020-03-15
Disable static checkers on generated js files

IagoLast:
This will work as a hotfix for #447

I think it could make sense to include it.
Since those are auto generated they never should be edited so it would be ok disable static checkers.

**Eslint:**
https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments
**Typescript**
https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -1013,7 +1013,9 @@ void PrintFileHeader(Printer* printer, const std::map<string, string>& vars) {
       " * @enhanceable\n"
       " * @public\n"
       " */\n\n"
-      "// GENERATED CODE -- DO NOT EDIT!\n\n\n");
+      "// GENERATED CODE -- DO NOT EDIT!\n\n\n"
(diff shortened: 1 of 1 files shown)


IagoLast:
@stanley-cheung happy to help! this is a very nice project ❤️ 

Going deeper into the causes of this issue is because the code relies on `googl` to import/export modules. Any plans to a ES6 modules version on the roadmap?

stanley-cheung:
No plan to support ES6 modules for now because protobuf doesn't support ES6 modules yet: https://github.com/protocolbuffers/protobuf/tree/master/js#protocol-buffers---googles-data-interchange-format

abjennings:
I'm still getting 'proto' is not defined (and 'COMPILED') on the protobuf JS file.

Do you think the protocolbuffers/protobuf project would add these lines to the protobuf JS file?

c10b10:
I can confirm this has only been added to the `--grpc-web_out` generated file. Could it please be added to the `--js_out` file as well?

PR #778, opened 2020-04-07
Make kokoro run under macOS

vbfox:
With this change kokoro download the `bazel` and `buildifier` binaries for mac when run under macOS


stanley-cheung:
Thanks for the contribution! This is very nice. I will get to this in a bit.

Issue #766, opened 2020-03-28
Interceptors implementation


stanley-cheung:
I just want to give an update here that we are working on the Interceptors implementation and should have something relatively soon. 

The overall design we are borrowing elements from #558 and have to take into consideration of our internal use cases as well so there might be some deviation from #558. One of the goals we have is to have a shared implementation of Interceptors both for our internal use case and the open source use case here. I am going to post some more details on the overall design, some examples here soon.

~~One of the challenges we faced while working on this, is that internally, our codebase is entirely written in Closure - so we can use directives like `@interface`, `@implements` to define the Interceptors interface. And generally the user app code which depends on our grpc-web closure library are compiled  _together_ with our grpc-web library. One of the issues we faced is that - since we attempt to define the `Interceptor` class as an `@interface` in Closure, all the methods got reduced away to something like `.a` when we compile the library into the npm module using `--compilation_level=ADVANCED_OPTIMIZATIONS`. There doesn't seem to be a good way for the open source use case here, to be able to write a custom interceptor while it `@implements` the `Interceptor` interface we want.~~

~~One workaround we found so far is to relax compiling our npm module to be using `--compilation_level=SIMPLE_OPTIMIZATIONS` instead. This will work - all our "interfaces" classes will survive the compilation, but the drawback is that the `index.js` file size of the module will increase significantly. We are not sure how much that will affect the final code size after using tools like `webpack` - we are still looking into that.~~

~~This all kind of stems from the fact that Closure and CommonJS has a very different style of exporting a module and we are still trying to figure out what's the best way to workaround this `@interface` issue while trying to maintain the same implementation both internally and externally (this will really keep us maintainers sane!).~~

Good news: we figured out how to export interfaces while building the module with ADVANCED_OPTIMIZATIONS, using `@export` and externs. 

So stay tuned - I am going to be posting more development on this in this issue soon. Thanks.

jmzwcn:
great

stanley-cheung:
Here's a high level view of what we are working on - these are still heavily WIP so things might change. We are still working through some issues with integrating with the internal use case, but we felt that this is a good checkpoint to write a summary.

---

In order to write your own gRPC-Web client interceptor, you will need to implement the following interface:

```
Interceptor.prototype.intercept = function(request, invoker) {};
```

[...]

PR #785, opened 2020-04-20
grpc-web interceptors implementation

stanley-cheung:
This PR introduces interceptors for grpc-web according to the design we spec'ed out in #766


clehene:
Just noted this was merged. That's really great! 
I notice it's missing the actual documentation. How are users going to know about it and once they do know how to use it?

stanley-cheung:
@clehene Yes good point. We have added an example using the interceptor here: https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/commonjs-example/client.js#L36. And we are working on a blog post. But to add to that, we probably should add something to our main README to point out some high level details about interceptors. Yes we should get to that in a bit.

PR #790, opened 2020-04-27
Fix typos in the in-process-proxy docs

jtattermusch:


jtattermusch:
@stanley-cheung  @wenbozhu

PR #851, opened 2020-06-10
Add a ThenableCall base class for the promise-based unaryCall function

stanley-cheung:
Driven by an internal refactor

author:@Jennnnny


stanley-cheung:
Internal change submitted. Merging here.

PR #854, opened 2020-06-15
Fix helloworld example docker command

kohashi:
@stanley-cheung 's update is really great!
https://github.com/grpc/grpc-web/pull/852

But the example Docker command didn't have port mapping, therefore it's not work in my environment (and probably anyone else's).


stanley-cheung:
I distinctly remember that running with `--network=host` on my linux machine gives a warning that those -p parameters will be ignored. And running with just `--network=host` on my linux machine was just fine. So I am inclined to leave it at that. Perhaps we just fix Line 305 (which is in a quote block), but leave L297 alone? Thanks for understanding.

kohashi:
Thanks for the message.  I hadn't tried it on a linux machine, but I needed the `-p` parameters in my Mac environment. I update commit (fix only L305).
Thanks for the kind message.

PR #859, opened 2020-06-20
Fix envoy.yaml deprecated fields

dmaixner:
This gets rid of warnings in envoy log files for deprecated fields.


stanley-cheung:
Thanks for the fix! Running tests.

PR #861, opened 2020-06-21
Separate Core code from Examples and interop-test code.

vnorigoog:
This CL is bringing this code to be ready for dogfooding with alpha customers.
Except for Readme files and some minor cleanup, this is now ready for alpha-dogfooding.


net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/proto/greeter.proto
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+syntax = "proto3";
+
+option java_package = "grpcweb.examples.greeter";
+
+package grpcweb.examples.greeter;
+
+// The greeting service definition.
+service Greeter {
```
stanley-cheung, line 21: There is a copy here: https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/helloworld/helloworld.proto (without the `package`). Is there someway we can reuse that copy of the `helloworld.proto`?

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
```diff
@@ -0,0 +1,106 @@
...
+import java.nio.ByteBuffer;
+import java.util.logging.Logger;
+import org.apache.http.HttpVersion;
+import org.apache.http.client.fluent.Request;
+import org.apache.http.entity.ContentType;
+
+/**
+ * A simple client that requests a greeting from the {@link GreeterService}.
+ */
+public class GreeterClient {
```
stanley-cheung, line 35: I am confused about the purpose of this - shouldn't the GreeterClient just be a simple program setting the proto message to 'foo', call the SayHello method, and print out the response? Why all these byte level validation code here?

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
```diff
@@ -0,0 +1,106 @@
...
+    byte[] packagedBytes = packageReqObj(reqObj);
+
+    // send request to the grpc-web server
+    ContentType contentType = ContentType.create("application/grpc-web");
+    byte[] response = Request.Post("http://localhost:" + proxyStarter.getGrpcwebPortNum() +
+        "/grpcweb.examples.greeter.Greeter/SayHello")
+        .useExpectContinue()
+        .version(HttpVersion.HTTP_1_1)
+        .bodyByteArray(packagedBytes, contentType)
+        .execute().returnContent().asBytes();
```
stanley-cheung, line 102: It seems confusing to have to do all these work to make a "grpc-web client in Java". Might be sending a wrong message here - we aren't trying to make a "grpc-web client in Java" here. This is purely for testing right?

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterService.java
```diff
@@ -0,0 +1,31 @@
...
+
+package grpcweb.examples.greeter;
+
+import grpcweb.examples.greeter.GreeterOuterClass.HelloReply;
+import grpcweb.examples.greeter.GreeterOuterClass.HelloRequest;
+import io.grpc.stub.StreamObserver;
+
+public class GreeterService extends GreeterGrpc.GreeterImplBase {
+  @Override
+  public void sayHello(HelloRequest req, StreamObserver<HelloReply> responseObserver) {
```
stanley-cheung, line 25: Yea this makes sense to me - a Java server.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/StartServiceAndGrpcwebProxy.java
```diff
@@ -0,0 +1,51 @@
...
+    LOG.info("grpc-port = " + grpcPort + ", grpc-web-port = " + grpcWebPort);
+
+    // Start the Grpc service on grpc-port
+    startGrpcService(grpcPort);
+
+    // Start the grpc-web proxy on grpc-web-port.
+    (new JettyWebserverForGrpcwebTraffic(grpcWebPort)).start();
+    // grpc-web proxy needs to know the grpc-port# so it could connect to the grpc service.
+    GrpcPortNumRelay.setGrpcPortNum(grpcPort);
+  }
```
stanley-cheung, line 50: This also makes sense to me - this is how you would build and start the Java GrpcWeb proxy

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
stanley-cheung: I see. You need a "grpc-web" client here. And this is quite a verbose way to do one in Java.

Let's discuss how we should approach here - I think making a "grpc-web client in Java" might have sent a wrong message here. We should probably re-use the JS client portion of this helloworld example here https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/helloworld as the client. Take out all the Envoy and the grpc server part, and replace with yours. But we should be able to re-use the JS client. That would fit the purpose of this repo better.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/proto/greeter.proto
vnorigoog: Yes it is duplicate.
add a comment to to it to say it is a copy.

It should be possible to reuse the same proto but maven is not allowing me to refer to files in parent directories. Have to figure that out.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
vnorigoog: > I am confused about the purpose of this - shouldn't the GreeterClient just be a simple program setting the proto message to 'foo', call the SayHello method, and print out the response? Why all these byte level validation code here?

You are right. We don't need the validation code.
I was just trying to show how to have a simple java client.

The README.md file will say
1. To start the server which includes proxy for grpc-web service, do the following
    mvn exec:java -Dexec.mainClass="grpcweb.examples.greeter.StartServiceAndGrpcwebProxy" test

2. and, then to test the greeter app, either use the javascript (you mentioned) or use a sample java client like this:
   mvn exec:java -Dexec.mainClass="grpcweb.examples.greeter.GreeterClient" test

3. Or, you could the following to sanity check the greeter service and client by doing this (which starts the service and client, runs the client, validates the response and then shuts everything down)
   mvn exec:java -Dexec.mainClass="grpcweb.examples.greeter.GreeterClientTest" test

Or, if it is confusing, we can completely get rid of GreeterClient.java and GreeterClientTest.java

what do you think?

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
vnorigoog: yes just for testing.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterService.java
vnorigoog: yes - just starts the server

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/StartServiceAndGrpcwebProxy.java
vnorigoog: yes just to show an example way to build the proxy into the customer's own service.

vnorigoog:
Thanks for the review. PTAL

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
stanley-cheung: I think this is fine as-is then. We can tweak this when we are working on the main README to put the right focus on things. We can just put a note here to say this is for testing only.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/proto/greeter.proto
stanley-cheung: I see. Yea we can improve this through iterations. Ideally we should avoid duplicating this.

stanley-cheung:
> Thanks for the review. PTAL

Let me know when you feel this is ready to run tests and merge. Thanks.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/proto/greeter.proto
vnorigoog: adding a TODO.

net/grpc/gateway/examples/grpc-web-java/greeter-service/src/main/java/grpcweb/examples/greeter/GreeterClient.java
vnorigoog: added comments to say this is just for testing purposes.

vnorigoog:
Thanks for the review. I will upload the latest with some TODOs to address the comments.

This is ready for test and merge

stanley-cheung:
@vnorigoog There seems to be conflicts.

PR #869, opened 2020-06-24
add scripts etc to run interop-tests on grpc-web java connector code

vnorigoog:


scripts/run_java_interop_tests.sh
```diff
@@ -0,0 +1,45 @@
...
+done
+
+# Build & Start the container with the grpc-web java service and proxy
+docker-compose build java-interop-server
+pid=$(docker run -d --network=host grpcweb/java-interop-server)
+
+sleep 15
+
+# Run interop tests
+docker run --network=host --rm grpcweb/prereqs /bin/bash /github/grpc-web/scripts/docker-run-interop-tests.sh
```
stanley-cheung, line 39: This is nice you can run exactly the same interop client!!

scripts/run_java_interop_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Build & Start the container with the grpc-web java service and proxy
+docker-compose build java-interop-server
+pid=$(docker run -d --network=host grpcweb/java-interop-server)
+
+# Run interop tests
+docker run --network=host --rm grpcweb/prereqs /bin/bash /github/grpc-web/scripts/docker-run-interop-tests.sh
+
+docker rm -f "$pid"
+
```
stanley-cheung, line 50: Can you put L42 - L49 above in here: https://github.com/grpc/grpc-web/blob/master/scripts/run_interop_tests.sh#L49

Then you don't need this `run_java_interop_tests.sh` script.

You can compile your L43 above with https://github.com/grpc/grpc-web/blob/master/scripts/run_interop_tests.sh#L34

vnorigoog: good idea! 
done.
PTAL

scripts/run_interop_tests.sh
```diff
@@ -47,6 +47,14 @@ docker rm -f "$pid1"
 docker rm -f "$pid2"
 
 
+#
+# Run interop tests against grpc-web java connector code
+#
+docker-compose build java-interop-server
```
stanley-cheung, line 53: Nit: can we merge this line with L34 above? So we build all the docker images first.

scripts/run_interop_tests.sh
```diff
@@ -47,6 +47,14 @@ docker rm -f "$pid1"
...
+pid3=$(docker run -d --network=host grpcweb/java-interop-server)
+docker run --network=host --rm grpcweb/prereqs /bin/bash /github/grpc-web/scripts/docker-run-interop-tests.sh
```
stanley-cheung, line 55: Nit: can we break this long line like L43-44 above? Just to be consistent.

Thanks

scripts/run_interop_tests.sh
vnorigoog: done

scripts/run_interop_tests.sh
vnorigoog: done. (made a shell function to call tests)

vnorigoog:
Thanks for the reviews. PTAL

stanley-cheung:
> Thanks for the reviews. PTAL

Thanks! Looks good. One final nit: can you please squash this down to maybe 1 commit please?

Issue #507, opened 2019-03-21
Migrate Bazel rule to toolchain


Yannic:
I'm planning to change the implementation of `closure_grpc_web_library` to use Bazel's support for [toolchains](https://docs.bazel.build/versions/master/toolchains.html) instead of hard-coding the compiler and runtime. This will give users more flexibility in how they can configure and use that rule.
For example: Assuming Google Open-Sources their JSON-like wire format or one of those alternative modes ([1](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpc_generator.cc#L49)), if someone wants to switch to that, they have to update every `closure_grpc_web_library`-target. When the configuration is passed as a toolchain, it's sufficient to tell Bazel to use that new toolchain, and updating individual targets is no longer required.

On the downside, this might be considered to be a breaking change because it will require a trivial (but manual) addition to the `WORKSPACE` file to register the (default) toolchain.
See [rules_go](https://github.com/bazelbuild/rules_go#setup) for an example of what that would look like.

Let me know if there are any concerns regarding this change.

Yannic:
Additionally, I'd like to propose that we ask for grpc-web to be added to Bazel's [CI system](https://github.com/bazelbuild/continuous-integration). That way, we will be notified if there are any breaking changes instead of periodically checking whether everything is still fine (#501, #494).
@stanley-cheung WDYT?

Yannic:
Since there hasn't been any feedback on this, I'll assume that everyone is happy with the changes proposed here.
I'll create a PR implementing this soon (TM).

sampajano:
Hi :) Closing due to being obsolescence — `closure_grpc_web_library` was cleaned up #1138. Thanks for the contributions, Yannic :)

PR #872, opened 2020-06-26
[bazel] Introduce grpc_web_toolchain

Yannic:
This change adds a new rule `grpc_web_toolchain` which we will use to
provide the generator/runtime to `closure_grpc_web_library`.

Updates #507

RELNOTES: [bazel] `closure_grpc_web_library` now uses toolchains to
resolve the generator and runtime. To migrate, add the following snipped
to your `WORKSPACE`:
```starlark
load("@com_github_grpc_grpc_web//bazel:repositories.bzl", "grpc_web_toolchains")
grpc_web_toolchains()
```

RELNOTES: [bazel] gRPC-Web now follows best practices for exporting
rules and exports all public symbols in `@com_github_grpc_grpc_web//bazel:defs.bzl`.
We intend to remove the existing `closure_grpc_web_library.bzl` in a
future release.


Yannic:
This is not ready for merging because the required change in Bazel has not been released yet. Requires Bazel@HEAD (> 3.3.x).

Yannic:
This is ready for review now. PTAL, thanks!

bazel/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+toolchain_type(
+    name = "toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+grpc_web_toolchain(
+    name = "closure_toolchain_impl",
+    generator = "//javascript/net/grpc/web:protoc-gen-grpc-web",
+    runtime_library = "//javascript/net/grpc/web:closure_grpcweb_runtime",
+)
```
stanley-cheung, line 27: I guess my question here is that, do the attribute `generator` and `runtime_library` have special meaning here? Or it can be any string?

Yannic: Both attributes are Labels that have to point to targets (just like, e.g, `cc_library`'s `deps`). `generator` also must be an executable target (e.g. `{cc,go,java}_binary`) and `runtime_library` must point to a `closure_js_library` (compatible) target.

PR #879, opened 2020-07-02
Add a 'goog_promise' option to generator

stanley-cheung:
Internal change

author:@Jennnnny


Yannic:
@stanley-cheung Using this requires a type `grpc.web.promise.GrpcWebPromise` that doesn't exist (yet). Is this intentional or was it an oversight?

stanley-cheung:
@Yannic This is supposed to be an internal feature. I thought that if we don't do anything, the `grpc.web.promise.GrpcWebPromise` will not be included in the generated file? I will double check tomorrow.

PR #884, opened 2020-07-07
Update protoc plugin to support Proto3 optional

stanley-cheung:
Proto3 introduced the `optional` syntax to support [field presence](https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md) tracking in [release 3.12.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.12.0). We need to add support in our `protoc-gen-grpc-web` protoc plugin.

See [this doc](https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md#signaling-that-your-code-generator-supports-proto3-optional) for more details. Basically we need to add a function `GetSupportedFeatures()` to our grpc protoc plugins to mark our compliance.

See similar work being done in the main grpc repo: https://github.com/grpc/grpc/pull/22998

This requires adding some bazel stuff to pin `@com_google_protobuf` to a higher version (i.e. at least `3.12.0` or above).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -11,8 +11,32 @@ http_archive(
     ],
 )
 
+http_archive(
+    name = "bazel_skylib",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.0.2/bazel-skylib-1.0.2.tar.gz",
+        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.0.2/bazel-skylib-1.0.2.tar.gz",
+    ],
+    sha256 = "97e70364e9249702246c0e9444bccdc4b847bed1eb03c5a3ece4f83dfe6abc44",
+)
+
+http_archive(
+    name = "com_google_protobuf",
+    sha256 = "efaf69303e01caccc2447064fc1832dfd23c0c130df0dc5fc98a13185bb7d1a7",
+    strip_prefix = "protobuf-678da4f76eb9168c9965afc2149944a66cd48546",
+    urls = [
+        "https://github.com/google/protobuf/archive/678da4f76eb9168c9965afc2149944a66cd48546.tar.gz",
+    ],
+)
+
+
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -11,8 +11,32 @@ http_archive(
...
+
+
 load("@io_bazel_rules_closure//closure:repositories.bzl", "rules_closure_dependencies", "rules_closure_toolchains")
 
 rules_closure_dependencies()
 
 rules_closure_toolchains()
+
+
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
```
Yannic, line 40: Not sure if this and the updated `bazel_skylib` is necessary.

Also, I think this file needs formatting (CI passes though, maybe we're not checking that WORKSPACE is formatted?).

stanley-cheung: Let me double check. Thanks for checking in - was going to copy you for review. 

I need to add the bazel_skylib because I got an error with something like `common_settings.bzl` not found (I lost that error message already).

stanley-cheung: And yes we probably only run `buildifier` on `BUILD.bazel` files but not `WORKSPACE` file. I will double check that too.

stanley-cheung: - Removed this `protobuf_deps` section. Confirmed there's no need for that.
- The error I got if I don't include this `bazel_skylib` is
```
ERROR: /usr/local/google/home/stanleycheung/grpc-web/javascript/net/grpc/web/BUILD.bazel:4:1: error loading
package '@com_google_protobuf//': Unable to load file '@bazel_skylib//rules:common_settings.bzl': file doesn't exist
and referenced by '//javascript/net/grpc/web:protoc-gen-grpc-web'
ERROR: Analysis of target '//javascript/net/grpc/web:protoc-gen-grpc-web' failed; build aborted: error loading package
'@com_google_protobuf//': Unable to load file '@bazel_skylib//rules:common_settings.bzl': file doesn't exist
```
- Ran `buildifier` against the `WORKSPACE` file as well now.

@Yannic, PTAL, thanks!

PR #886, opened 2020-07-08
Add missing types definitions

### LINKED ISSUE #848: Update ts typings for support interceptors
https://github.com/grpc/grpc-web/issues/766

https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/index.d.ts

### LINKED ISSUE #868: Add MethodDescriptor to TypeScript definitions
A followup of https://github.com/grpc/grpc-web/issues/865#issuecomment-649000227

The TypeScript definitions do not have the `MethodDescriptor` exported.

With bypassing the typings

```ts
declare var require;

console.log(require('grpc-web').MethodDescriptor);
```

I can read it in TypeScript.

Looks like grpc-web's TypeScript mode still uses the deprecated `MethodInfo` instead and should be updated as well.

### LINKED ISSUE #877: UnaryInterceptor not called
I guess this is going something forgotten on my side, but after spending some time with this I have to ask.

### Description

Backend is in Go. Using Envoy for proxy. Svelte for frontend.

My proto file is pretty simple:
```javascript
...
message HelloResponse {
    string name = 1;
}

message IdRequest {
    int32 id = 1;
}

service Backend {
    rpc GetHello (IdRequest) returns (HelloResponse) {}
}
```

I have used updated [unary interceptor from blog post](https://grpc.io/blog/grpc-web-interceptor/#unary-interceptor-example):
```javascript
  const SimpleUnaryInterceptor = function () {};
  SimpleUnaryInterceptor.prototype.intercept = function (request, invoker) {
    console.log("test1");
    const reqMsg = request.getRequestMessage();
    reqMsg.setId("[Intercept request]" + reqMsg.getId());
    return invoker(request).then((response) => {
      console.log("test2");
      const responseMsg = response.getResponseMessage();
      responseMsg.setName("[Intercept response]" + responseMsg.getName());
      return response;
    });
  };
```

I have bound it to the client via opts:
```javascript
  let opts = { 'unaryInterceptors': [new SimpleUnaryInterceptor()] };
  let client = new protobuf.BackendClient("https://myserver:8092", null, opts);
```

I have this simple request/response code:
```javascript
  let request = new protobuf.IdRequest();
  request.setId(3);
  client.getHello(request, {}, (err, response) => {
    console.log(response.getName());
  });
```

### Problem
Request and response work fine, but interceptor doesn't seem to be run at all. In console log I can see only response data from server, but nothing from interceptor. Can someone please tell me, what am I missing here?

Thank you.

stanley-cheung:
Fixes #848: Fixed a bug where we can't pass the interceptors into the client constructor because of a type mismatch.

Fixes #868: Added `MethodDescriptor` to the exported types. Replace the deprecated `MethodInfo` with `MethodDescriptor`.

Fixes #877: Fixed issue where `UnaryResponse` symbols are being optimized away.

And in general added some missing classes to the exported types as well.

Added tests for both callback-based `StreamInterceptor` and promise-based `UnaryInterceptor`. And now execute `tsc` with `--strict`.

This is WIP.

### DIFF
--- a/packages/grpc-web/index.d.ts
+++ b/packages/grpc-web/index.d.ts
@@ -3,47 +3,50 @@ declare module "grpc-web" {
   export interface Metadata { [s: string]: string; }
 
   export namespace AbstractClientBase {
-    class MethodInfo<Request, Response> {
-      constructor (responseType: new () => Response,
-                   requestSerializeFn: (request: Request) => {},
-                   responseDeserializeFn: (bytes: Uint8Array) => Response);
+    class MethodInfo<REQ, RESP> {
+      constructor (responseType: new () => RESP,
+                   requestSerializeFn: (request: REQ) => {},
+                   responseDeserializeFn: (bytes: Uint8Array) => RESP);
     }
   }
 
   export class AbstractClientBase {
-    unaryCall<Request, Response> (method: string,
-             request: Request,
-             metadata: Metadata,
-             methodInfo: AbstractClientBase.MethodInfo<Request, Response>
-            ): Promise<Response>;
+    thenableCall<REQ, RESP> (
+      method: string,
+      request: REQ,
+      metadata: Metadata,
+      methodDescriptor: MethodDescriptor<REQ, RESP>
+    ): Promise<RESP>;
 
-    rpcCall<Request, Response> (method: string,
-             request: Request,
-             metadata: Metadata,
-             methodInfo: AbstractClientBase.MethodInfo<Request, Response>,
-             callback: (err: Error, response: Response) => void
-            ): ClientReadableStream<Response>;
+    rpcCall<REQ, RESP> (
+      method: string,
+      request: REQ,
+      metadata: Metadata,
+      methodDescriptor: MethodDescriptor<REQ, RESP>,
+      callback: (err: Error, response: RESP) => void
+    ): ClientReadableStream<RESP>;
 
-    serverStreaming<Request, Response> (method: string,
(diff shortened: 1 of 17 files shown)


packages/grpc-web/index.d.ts
```diff
@@ -32,31 +32,71 @@ declare module "grpc-web" {
...
                     
     cancel (): void;
   }
 
+  export interface StreamInterceptor<Req, Resp> {
+    intercept(request: Request<Req, Resp>,
+              invoker: (request: Request<Req, Resp>) =>
+      ClientReadableStream<Resp>): ClientReadableStream<Resp>;
+  }
+
```
willsalz, line 69: Thanks for putting this up! I was wondering if you could also add UnaryInterceptor while you're at it.

From: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/interceptor.js#L47

```suggestion

  export interface UnaryInterceptor<Req, Resp> {
      intercept(request: Request<Req, Resp>,
                invoker: (request: Request<Req, Resp>) =>
        Promise<!UnaryResponse<Resp>>): Promise<UnaryResponse<Resp>>;
    }
  
```

stanley-cheung: Yes literally working on that part as we speak :)

stanley-cheung: @willsalz Added the `UnaryInterceptor` interface typings to the PR. Still WIP. Hopefully should land soon.

PR #898, opened 2020-07-14
Update Bazel to 3.4.1

Yannic:


Yannic:
This should unblock #872 (which, of course, needs rebasing after this lands)

PR #913, opened 2020-07-21
Update MethodDescriptorInterface

stanley-cheung:
Exporting an internal change

author:@Jennnnny


stanley-cheung:
Internal change merged.

PR #915, opened 2020-07-23
missing decode grpc error message

wapa5pow:
Like https://github.com/grpc/grpc-web/blob/5276b1e36deabb64776c9af6bfe6be5f83fa1ba2/javascript/net/grpc/web/grpcwebclientreadablestream.js#L212 , grpc-message should be decoded by decodeURIComponent.


wapa5pow:
Actually, `sendErrorCallbacks_` is not decoded. I refactored code.

PR #929, opened 2020-08-17
Update envoy.yaml

dwaiba:
Changed deprecated option 'envoy.api.v2.route.CorsPolicy.allow_origin' to
```
                allow_origin_string_match:
                  - safe_regex:
                     google_re2: {}
                     regex: \*
```

### DIFF
--- a/net/grpc/gateway/examples/helloworld/envoy.yaml
+++ b/net/grpc/gateway/examples/helloworld/envoy.yaml
@@ -27,7 +27,9 @@ static_resources:
                   max_grpc_timeout: 0s
               cors:
                 allow_origin_string_match:
-                - prefix: "*"
+                  - safe_regex:
+                     google_re2: {}
+                     regex: \*
                 allow_methods: GET, PUT, DELETE, POST, OPTIONS
                 allow_headers: keep-alive,user-agent,cache-control,content-type,content-transfer-encoding,custom-header-1,x-accept-content-transfer-encoding,x-accept-response-streaming,x-user-agent,x-grpc-web,grpc-timeout
                 max_age: "1728000"


net/grpc/gateway/examples/helloworld/envoy.yaml
```diff
@@ -27,7 +27,9 @@ static_resources:
                   max_grpc_timeout: 0s
               cors:
                 allow_origin_string_match:
-                - prefix: "*"
```
stanley-cheung, line 30: Can you point us to where this `prefix` field is being deprecated? I still see it here in the docs: https://www.envoyproxy.io/docs/envoy/latest/api-v2/type/matcher/string.proto#envoy-api-msg-type-matcher-stringmatcher.

dwaiba: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto.html?highlight=allow_origin#route-corspolicy states that allow_origin field has been deprecated in favor of allow_origin_string_match

stanley-cheung: But we are already using `allow_origin_string_match`, right?

stanley-cheung: This is already done back in #716

dwaiba: Thanks.  That was in the echo example but the helloworld was unchanged.

stanley-cheung: But the helloworld/envoy.yaml _is_ changed to be using `allow_origin_string_match` already right?

dwaiba: My Bad... Generally use `safe_prefix `with `allow_origin_string_match` . Closing.

dwaiba:
My Bad... Generally use` safe_prefix` with `allow_origin_string_match `. Closing.

PR #941, opened 2020-08-28
Fix Protobuf .d.ts typings for .proto files without package

### LINKED ISSUE #939: Protocol buffer enum first character dropped in grpc_generator.cc
I noticed that in a proto file when there is no `package` set, then the enums are missing the first character.

```
syntax = "proto3";

enum Status {
    CREATED = 0;
    PROCESSING = 1;
}

message InfoResponse {
    Status status = 1;
}
```

will produce   **`tatus` instead of `Status`**

```

export class InfoResponse extends jspb.Message {
  getStatus(): tatus;
  setStatus(value: tatus): InfoResponse;

  toObject(includeInstance?: boolean): InfoResponse.AsObject;
  static toObject(includeInstance: boolean, msg: InfoResponse): InfoResponse.AsObject;
  static serializeBinaryToWriter(message: InfoResponse, writer: jspb.BinaryWriter): void;
  static deserializeBinary(bytes: Uint8Array): InfoResponse;
  static deserializeBinaryFromReader(message: InfoResponse, reader: jspb.BinaryReader): InfoResponse;
}

export namespace InfoResponse {
  export type AsObject = {
    status: tatus,
  }
}

export enum Status { 
  CREATED = 0,
  PROCESSING = 1,
}
```

It can be fixed by setting a `package` but it worked fine with empty package with version 1.1.0

I think this change is responsible:
https://github.com/grpc/grpc-web/commit/cd059a28b274ca9c1ffe38f725a91a294afa5797#diff-2e729f9088e537bfeb6fefc318963a96

` return enum_name.substr(1);` instead of 
```
if (!js_field_type.empty() && js_field_type[0] == '.') {
      js_field_type = js_field_type.substr(1);
    }
```

Yannic:
Fixes #939

### DIFF
--- a/javascript/net/grpc/web/grpc_generator.cc
+++ b/javascript/net/grpc/web/grpc_generator.cc
@@ -367,11 +367,13 @@ string JSElementType(const FieldDescriptor *desc, const FileDescriptor *file) {
 
     case FieldDescriptor::TYPE_ENUM:
       if (desc->enum_type()->file() == file) {
+        // [for protobuf .d.ts files only] Do not add the module prefix for
+        // local messages.
         string enum_name =
             StripPrefixString(
                 desc->enum_type()->full_name(),
                 desc->enum_type()->file()->package());
-        return enum_name.substr(1);
+        return StripPrefixString(enum_name, ".");
       }
       return ModuleAlias(desc->enum_type()->file()->name())
           + StripPrefixString(
--- a/packages/grpc-web/test/protos/test01.proto
+++ b/packages/grpc-web/test/protos/test01.proto
@@ -5,5 +5,15 @@ message MessageOuter {
     int32 value = 1;
   }
 
+  enum EnumInner {
+    DEFAULT = 0;
+  }
+
   repeated MessageInner someProp = 1;
+  EnumOuter someEnum = 2;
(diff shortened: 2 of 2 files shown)


stanley-cheung:
@uwemaurer Thanks for the report.
@Yannic Thanks for the fix.

I was wondering whether we can add some tests for this. 

There were some tests involving stuff with no proto package before. See this [test proto](https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/protos/nopackage.proto), and [this test](https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/plugin_test.js#L170) for trying to execute these relevant code path in the codegen plugin.

If this has to do with Dts generation and we need to test whether some generated code survives the `tsc` compilation, we have some proto like [this](https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/protos/test01.proto), such that we have tests like [this](https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/tsc_test.js#L59) first to run the plugin to generate the generated code (which would exercise the code changed in for example this PR), and then run `tsc` on the generated code to make sure things are compiled properly.

Yannic:
@stanley-cheung done! ptal, thanks.

PR #964, opened 2020-09-24
Make grpc-web Java example code to be more like a unittest - so that …

vnorigoog:
…we can easily run it with maven command.


stanley-cheung (approved):
LGTM

PR #1026, opened 2021-01-22
TypeScript declarations for MethodType and MethodDescriptor

pro-wh:
We'd like to use these in our ts code, where we're not using the generator.


pro-wh:
thanks!

PR #1041, opened 2021-02-12
Update references to internal targets to use workspace root

### LINKED ISSUE #1015: Incorrect path in WORKSPACE
The path to the target "//bazel:closure_toolchain" should be scoped under "@com_github_grpc_grpc_web" 

https://github.com/grpc/grpc-web/blob/e88fdc809d9ccbf1c9dae701ee14cce53be38b9e/bazel/repositories.bzl#L26

WillEngFlow:
This solves issues when this project is used as an external dependency
in other workspaces. Previously Bazel would not be able to find these
resources.

This fixes #1015

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -44,7 +44,7 @@ rules_closure_dependencies()
 
 rules_closure_toolchains()
 
-load("//bazel:repositories.bzl", "grpc_web_toolchains")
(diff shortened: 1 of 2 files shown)


stanley-cheung:
cc @Yannic

Yannic (changes requested):
Resolved this with Will offline, these changes aren't needed.

PR #1057, opened 2021-03-13
Set upstream timeout for envoy in echo example

Yannic:
Without this, streaming calls fail after a few seconds because envoy terminates the connection to the gRPC upstream.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-timeout


stanley-cheung (approved):
LGTM. Thanks for the contribution!

Issue #1076, opened 2021-05-02
If an empty `.proto` is specified to `closure_grpc_web_library()` it will fail.


psigen:
It looks like the `closure_grpc_web_library()` bazel rule might not quite handle empty proto files correctly.

I have created a mostly empty `.proto` file like so:

**order.proto**
```protobuf
syntax = "proto3"
# No other content in this file.
```

**BUILD.bazel**
```bazel
load("@com_github_grpc_grpc_web//bazel:defs.bzl", "closure_grpc_web_library")
load("@rules_proto//proto:defs.bzl", "proto_library")

proto_library(
    name = "proto",
    srcs = ["order.proto"],
)

closure_grpc_web_library(
    name = "web",
    import_style = "es6",
    mode = "grpcwebtext",
    deps = [":proto"],
)
```

When I run this, I get the following output:

```bash
$ bazel build //:web --verbose_failures
INFO: Invocation ID: 0f8abf94-0dd8-4e27-9b06-1d39d9e85295
INFO: Analyzed target //:web (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
INFO: From Generating GRPC Web for //:web:
ERROR: /home/pras/testrepo/BUILD.bazel:40:25: output 'order_grpc_web_pb.js' was not created
ERROR: /home/pras/testrepo/BUILD.bazel:40:25: output 'order.pb.grpc-web.js' was not created
ERROR: /home/pras/testrepo/BUILD.bazel:40:25: Generating GRPC Web for //:web failed: not all outputs were created or valid
Target //:web failed to build
INFO: Elapsed time: 0.418s, Critical Path: 0.26s
INFO: 2 processes: 1 remote cache hit, 1 internal.
FAILED: Build did NOT complete successfully
```

I think this is because `protoc` does not create output for these blank files, but the bazel rule still declares them as a target.  A simple solution might be to `touch` the files to create empty placeholders before running `protoc`.

psigen:
An interesting indirectly related conversation about this topic on the python protoc generator.
https://github.com/grpc/grpc/issues/14172

psigen:
[Here is the problem area of the code.](https://github.com/grpc/grpc-web/blob/master/bazel/private/rules/closure_grpc_web_library.bzl#L72-L88)  Each source file gets `actions.declare_file` called, but the [`protoc` call](https://github.com/grpc/grpc-web/blob/master/bazel/private/rules/closure_grpc_web_library.bzl#L95-L103)
[...]

PR #1077, opened 2021-05-02
Always generate files in closure_grpc_web_library()

psigen:
~This adds a `ctx.actions.write` to each of the automatically generated source targets that have been declared.~

This adds a flag `allow_empty_files` which forces the generation of output `.grpc.js` files even if they do not contain service definitions.  The flag is `False` by default, to retain the current behavior.  It is set `True` in the bazel plugin, and should be set when using any build system that tracks generated output files.

This prevents an issue if the source targets do not yield a `.grpc.js` generated file because they do not contain an `rpc` definition (e.g. proto files that only contain a `message`).

Resolves: #1076


psigen:
Thanks, looks like this will be a bit more complicated...

psigen:
#1085 is a better approach than what I was trying here.

PR #1085, opened 2021-05-22
Generate empty files in closure_grpc_web_library() bazel rules

psigen:
This adds a flag `allow_empty_files` which forces the generation of output `.grpc.js` files even if they do not contain service definitions.  The flag is `False` by default, to retain the current behavior.  It is set `True` in the bazel plugin, and should be set when using any build system that tracks generated output files.

This prevents an issue if the source targets do not yield a `.grpc.js` generated file because they do not contain an `rpc` definition (e.g. proto files that only contain a `message`).

Resolves: #1076
Supersedes: #1077

### DIFF
--- a/bazel/private/rules/closure_grpc_web_library.bzl
+++ b/bazel/private/rules/closure_grpc_web_library.bzl
@@ -59,6 +59,7 @@ def _generate_closure_grpc_web_srcs(
 
     args.add_all(_proto_include_paths(transitive_sources.to_list()), format_each = "-I%s")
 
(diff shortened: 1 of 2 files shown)


psigen:
@stanley-cheung this version passed all the tests locally.

sampajano:
Thanks a lot for the PR!

But unfortunately, we've decided to clean up our Bazel + javascript uses in https://github.com/grpc/grpc-web/pull/1138..

So this change no longer applies..

Sorry for the change!

PR #1088, opened 2021-06-01
add Metadata property to Error class.

tzmfreedom:
`Error` class has optional Metadata property. However, this is not implemented in type definition.
So type check is always failed whenever metadata is used.
This PR fixed Error class definition.


sampajano:
Thanks a lot for the PR! :)

In https://github.com/grpc/grpc-web/pull/1139 we've improved the error typing so the new `RpcError` class already includes the metadata field now! So I'll close this PR since it's no longer needed.

Thanks again for contribution tho! :)

PR #1107, opened 2021-07-17
Allow for custom install prefix

06kellyjac:
Allow for custom install prefix
Also create the dir if it doesn't exist and ensure 755 perms

### DIFF
--- a/javascript/net/grpc/web/Makefile
+++ b/javascript/net/grpc/web/Makefile
@@ -16,6 +16,7 @@ CXX ?= g++
 CPPFLAGS += -I/usr/local/include -pthread
 CXXFLAGS += -std=c++11
 LDFLAGS += -L/usr/local/lib -lprotoc -lprotobuf -lpthread -ldl
+PREFIX ?= /usr/local
 
 all: protoc-gen-grpc-web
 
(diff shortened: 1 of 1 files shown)


06kellyjac:
@stanley-cheung @sampajano Please can I get a review on this

This is required for us to easily package this for nixpkgs https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/tools/protoc-gen-grpc-web/default.nix#L16-L25

javascript/net/grpc/web/Makefile
```diff
@@ -16,14 +16,15 @@ CXX ?= g++
...
+PREFIX ?= /usr/local
 
 all: protoc-gen-grpc-web
 
 protoc-gen-grpc-web: grpc_generator.o
 	$(CXX) $^ $(LDFLAGS) -o $@
 
 install: protoc-gen-grpc-web
-	install protoc-gen-grpc-web /usr/local/bin/protoc-gen-grpc-web
+	install -Dm755 protoc-gen-grpc-web $(PREFIX)/bin/protoc-gen-grpc-web
```
sampajano, line 27: I don't think `-Dm755` work on macOS (BSD) version of `install`..

Is there a way you can make this work for both? :)

stanley-cheung: Can we leave this `-Dm755` out and just use the system default? the `PREFIX` change looks good.

06kellyjac: I can either drop it or add a `mkdir -p $(PREFIX)` above `install`

sampajano: Yeah this is cool with me.. But i'll defer to @stanley-cheung's for approval :)

Thanks for the change :)

stanley-cheung (approved):

PR #1116, opened 2021-08-04
Require assert in hello world README example

oflisback:
This is needed to make the example work, in the example code https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/helloworld/server.js the line is already included.


stanley-cheung:
@oflisback thanks for the contribution. Can you please complete the EasyCLA process please?

sampajano:
It seems that the EasyCLA process is completed so we're unblocked here? :)

sampajano (approved):
Thanks for the change! :)

PR #1131, opened 2021-09-01
Replace use of `third_party/closure-library` submodule

sampajano:
... in favor of `google-closure-library` npm package.

This is the first step in a series of efforts to streamline our Closure dependencies.

Benefit:
- More consistent closure library and compiler versions used across code examples and the npm release.

Notable changes:
- Removed `third_party/closure-library` in favor of using the `google-closure-library` npm package.
  - Started using `google-closure-library` in both our main npm package (`packages/grpc-web`) and the echo closure example. (Retained the existing version `v20201102` used in our npm release)
  - Removed the third_party/closure-library submodule following the instructions here: https://git.wiki.kernel.org/index.php/GitSubmoduleTutorial#Removal
- Updated the `net/grpc/gateway/docker/closure_client/` to use npm `closure-compiler` rather than a custom version of java closure compiler.
  - Also tweaked its Makefile and removed references to it in the main Makefile (since it doesn't run by itself)

Tested:
- Verified that `npm run build` inside `packages/grpc-web` generates exactly the same output as before.
- Test ran `closure-client` example: `docker-compose up --build node-server envoy closure-client`
- Test ran `commonjs-client` example: `docker-compose up --build node-server envoy commonjs-client`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "third_party/closure-library"]
-	path = third_party/closure-library
-	url = https://github.com/google/closure-library.git
 [submodule "third_party/protobuf"]
 	path = third_party/protobuf
 	url = https://github.com/protocolbuffers/protobuf.git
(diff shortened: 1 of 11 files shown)


sampajano:
Thanks for much for review, Stanley! 😃

Issue #1043, opened 2021-02-17
make a new release


pro-wh:
Maintainers, could you make a new release of grpc-web? We'd like to start using the fix from #1025.

stanley-cheung:
Acknowledged. Sorry about the delay.

JPeaker:
@stanley-cheung Any update on a new release coming? There's a specific bug that has been fixed in `main` that I'd really like to make use of!

jayars:
Hi, am also wondering if there's a new release on the roadmap?
Hoping to get the fix from https://github.com/grpc/grpc-web/issues/939.
Thanks

sampajano:
Thanks for the interest and sorry again for the delay!

We're definitely committed to cutting a new release soon.. It's just that there are currently some blocker issues (due to our dependencies) that we're working through.. But hopefully will be resolved soon :)

Thanks for the patience! :)

PR #1137, opened 2021-09-21
Revamp Closure JsUnit tests runtime and optimize test/build flows.

sampajano:
### Key changes:

* Removed dependency on [Bazel](https://bazel.build/) / [rules_closure](https://github.com/bazelbuild/rules_closure) for Closure JsUnit tests in favor of [google-closure-deps](https://www.npmjs.com/package/google-closure-deps) and [Protractor](https://www.protractortest.org/#/).
  * This allow us the direct use use most up-to-date version of [Closure Library](https://github.com/google/closure-library) (v.s. [indirectly used](https://github.com/bazelbuild/rules_closure/blob/321e678c64656f9dfc39e978a0a017d2930eb7e8/closure/repositories.bzl#L709-L720) through rules_closure), and unblock code sync and [new releases](https://github.com/grpc/grpc-web/issues/1043).
    * Also ensures **tests are ran on the same version of Closure as release** (which was not before).
  * Protractor setup uses generated test HTML files, and is HEAVILY inspired by [FirebaseUi](https://github.com/firebase/firebaseui-web) (mainly just Dockerizing their [solution](https://github.com/firebase/firebaseui-web/tree/master/buildtools)) :)

* Optimized Test flows
  * Restructured Dockerfile layouts and test scripts so that unit tests runs earlier and **much faster!**
    * e.g. when user makes a javascript changes, reruns will **fail in as early as 10 seconds** (on jsunit failures) and at most in ~60 seconds (on mocha test failures), **compared to 4+ minutes as before** ! (on 2019 MacbookPro)
  * `npm test` in `packages/grpc-web` now runs both JsUnit tests on top of Mocha tests.

* Optimized Dockerfiles
  * Removed `common` Docker images (saves disk space) and revamped `prereqs` to use  [multistage](https://docs.docker.com/develop/develop-images/multistage-build/) build.
    * Avoids the  possibility code from previous `git clone` to be accidentally in local builds
  * Optimized `prereqs` for faster build
    * Reduced build targets so `prereqs` build time is down from 5+ minutes to <3 minutes (on 2019 MacbookPro)
    * Optimized build sequence so larger / less frequently changed targets are built earlier to best leverage the Docker [build cache](https://pythonspeed.com/articles/docker-caching-model/).

### Other (small) changes:

* Fix small flake in interop test where it sometimes times out at 500ms (updated timeout to 1 second)
  * .. and updated mocha unit tests timeout to 10 seconds
* Cleaned up echo example Makefile to ignore testing files (rather than removing then adding them back)
* Created `.dockerignore` to avoid build failures when local node_modules/ are present
* Deleted `streambodyclientreadablestream.js` since it’s not used and doesn’t build.

### Follow-up:

* Clean up remaining `closure_js_library` usages and our [bazel rules](https://github.com/grpc/grpc-web/tree/master/bazel).


net/grpc/gateway/docker/node_interop_server/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM grpcweb/common
+FROM node:12-stretch
```
stanley-cheung, line 15: As discussed, perhaps let's lock the version here to `node:12.22.6-stretch` to match what we have in the `selenium` image below.

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -12,39 +12,54 @@
...
+
+RUN git clone https://github.com/grpc/grpc-web .
+COPY ./scripts/init_submodules.sh ./scripts/
+RUN ./scripts/init_submodules.sh
+
+
+######################################
+# Stage 2: Copy source files and build
+######################################
+FROM node:12-stretch AS copy-and-build
```
stanley-cheung, line 52: Same comment as above.

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -12,39 +12,54 @@
...
-RUN echo "\nloglevel=error\n" >> $HOME/.npmrc
+######################################
+# Stage 1: Fetch binaries
+######################################
+# node:12-stretch depends is based on buildpack-deps:stretch
+FROM buildpack-deps:stretch AS prepare
 
-WORKDIR /github/grpc-web
+ARG BUILDIFIER_VERSION=1.0.0
+ARG PROTOBUF_VERSION=3.17.3
```
stanley-cheung, line 25: Perhaps consider protobuf `3.18.0` they have just released.

packages/grpc-web/docker/jsunit-test/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM selenium/standalone-chrome:93.0
+
+# Node version in node:12-stretch image
+ARG NODE_VERSION=12.22.6
+
+USER root
+
+RUN apt-get update && \
+  apt-get install -y nodejs npm python
```
stanley-cheung, line 9: Please make sure this installs `python3`.

stanley-cheung (dismissed):
This is great work! Thanks for modernizing the build and test flows. Thanks for doing all the research into how other JS projects do this, distill and simplify the workflow to match our use case. This will unblock the effort to sync our codebase and work towards finally be able to make another release!

net/grpc/gateway/docker/node_interop_server/Dockerfile
sampajano: Very good idea! Done! :)

net/grpc/gateway/docker/prereqs/Dockerfile
sampajano: Done too :)

packages/grpc-web/docker/jsunit-test/Dockerfile
sampajano: Good catch! Removed python here since the selenium image already has Python 3.8 installed :)

net/grpc/gateway/docker/prereqs/Dockerfile
sampajano: Good idea! But since we need to update the git submodules too maybe doing a follow-up would be better :)

stanley-cheung (approved):

sampajano:
Thanks so much for the review, Stanley!! Happy to improve things! :)

PR #1082, opened 2021-05-12
Internal code sync

stanley-cheung:
author:@Jennnnny

PR #1138, opened 2021-09-22
Clean up Javascript-related Bazel rules.

sampajano:
### Summary

- Removes `closure_js_library` targets for they're no longer used for neither release or testing.
- Also removes `closure_grpc_web_library` and `grpc_web_toolchain`, for they depend on rules_closure, and hence an [older version of Closure](https://github.com/bazelbuild/rules_closure/blob/321e678c64656f9dfc39e978a0a017d2930eb7e8/closure/repositories.bzl#L709-L720) than what's [used](https://github.com/grpc/grpc-web/blob/32fe12459b02e6e8b3ed6f962c354bc2e5e09dfd/packages/grpc-web/package.json#L32) for release and testing.


stanley-cheung (approved):
Thanks for bringing this home. We are that much closer to being able to sync now!

stanley-cheung:
@Yannic FYI about this PR. 

We were having trouble (#1082) syncing code between our internal code base with Github here because `rules_closure` is embedding an outdated version of Closure.

@sampajano helped us migrate to be using Protractor to run the jsunit tests (#1137) instead of relying on bazel. Hence a lot of those closure bazel rules are no longer needed.

Since you have contributed a lot to this area in the past so we'd like to get your attention on this. Thank you so much contributing! And just to clarify, we will still be using bazel for other purposes in this repo (the cpp generator, etc). We only migrated away the piece where we were using bazel to run the closure jsunit tests away.

Thanks!

sampajano:
@Yannic Yeah thanks so much for contributions in the past.. I can see it's quite a lot of work.. And sorry for deleting it.. (But at least the code will be remembered forever by Github just in case we might need it one day in the future :P)

PR #1139, opened 2021-09-22
Internal code sync

sampajano:
### Summary
  * Internal code sync (up to Aug 4, 2021 for now..) & relevant improvements. :)
  * Updated Closure dependency to `20210808.0.0` (necessary dependency)

### Main code changes:
  * Improving error typing to use type `RpcError` in favor of untyped `Error`
    * Also updated `StatusCode` to be a enum (not full enum support (e.g. `RpcError[RpcError.OK] === 'OK'`) is there but general `===` comparison should work)
  * Removed `AbstractClientBase.MethodInfo`

### Other changes:
  * Unifying error behavior in TS and JS echo examples
  * Improved jsunit test runner to throw error when no unit test was found

### TODO:
  * https://github.com/grpc/grpc-web/pull/1063 is clobbered now so we need to redo it as a follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ const request = new EchoRequest();
 request.setMessage('Hello World!');
 
 const call = echoService.echo(request, {'custom-header-1': 'value1'},
-  (err: grpcWeb.Error, response: EchoResponse) => {
+  (err: grpcWeb.RpcError, response: EchoResponse) => {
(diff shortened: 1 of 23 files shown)


stanley-cheung (dismissed):
Yay! Finally we are able to do the sync!

stanley-cheung (approved):

PR #1140, opened 2021-09-23
Internal code sync

sampajano:
### Notable changes:

- Improve typing of `RpcError.code` from `number` -> `StatusCode` (Thanks @richieforeman! 😃)
- Returning `StatusCode.INTERNAL` during deserialization errors (per [spec](https://grpc.github.io/grpc/core/md_doc_statuscodes.html))
- Moved protoc generator to its own package


stanley-cheung (approved):
Thanks for doing this!

PR #1160, opened 2021-11-03
Expose MethodDescriptor's public methods

tomferreira:
Expose the public methods of the MethodDescriptor class to be used in a mock implementation proposal of interceptor.


sampajano (commented):
Thanks for the contrib! :)

This in principle is fine with me. I'm a bit surprised that these aren't exported already given it's already in the TS interface.

@stanley-cheung would probably have more context here so i'll wait for his review :)

(BTW you need to finish the CLA process for this PR to be merge-able :))

packages/grpc-web/externs.js
```diff
@@ -32,3 +32,11 @@ module.UnaryResponse.prototype.getResponseMessage = function() {};
...
 module.UnaryResponse.prototype.getMethodDescriptor = function() {};
 module.UnaryResponse.prototype.getStatus = function() {};
+
+module.MethodDescriptor = function() {};
+module.MethodDescriptor.getName = function() {};
+module.MethodDescriptor.getMethodType = function() {};
+module.MethodDescriptor.getResponseMessageCtor = function() {};
+module.MethodDescriptor.getRequestMessageCtor = function() {};
+module.MethodDescriptor.getResponseDeserializeFn = function() {};
+module.MethodDescriptor.getRequestSerializeFn = function() {};
```
stanley-cheung, line 42: Nit: please make sure file ends in a newline.

stanley-cheung (approved):
This LGTM. Yea `MethodDescriptor` should be part of the public API.

sampajano (approved):
Thanks again for the change! :)

sampajano:
Hi @tomferreira.. Unfortunately i have to rollback this PR because after merging this change it's causing some non-trivial code size increases for some internal Google products.. Apologize for the inconveniences.. 😅

I think maybe there are some ways to expose these APIs ONLY for Github. But I'm not totally sure yet. Ideas are welcome too.. 😃

PR #1164, opened 2021-11-08
Add missing class exports

### LINKED ISSUE #1158: Missing RpcError in exports.js (among other classes), inconsistent with index.d.ts
A relatively simple issue but I think quite important:
The `index.d.ts` contains definitions like the following (grpc-web = 1.3.0):

```
  export class UnaryResponse<REQ, RESP> {
    getResponseMessage(): RESP;
    getMetadata(): Metadata;
    getMethodDescriptor(): MethodDescriptor<REQ, RESP>;
    getStatus(): Status;
  }
  export class GrpcWebClientBase extends AbstractClientBase {
    constructor(options?: GrpcWebClientBaseOptions);
  }

  export class RpcError extends Error {
    constructor(code: StatusCode, message: string, metadata: Metadata);
    code: StatusCode;
    metadata: Metadata;
  }
```

There is nothing wrong with these definitions and they are very useful. However, `export`ing a class implies that I must be able to extend them, e.g. the following must work (and indeed compiles fine in Typescript):
```
class GRPCCallbackError extends RpcError {
  callId: string

  constructor (code: StatusCode, message: string, metadata: Metadata, callId: string) {
    super(code, message, metadata)
    this.callId = callId
  }
}
```

However, running this code throws an `TypeError: Class extends value undefined is not a constructor or null` because `index.js` of `grpc-web` does not actually export the `RpcError` class (or the `UnaryResponse` class for that matter). I confirmed this by checking `grpc-web/packages/grpc-web/exports.js` where only `GrpcWebClientBase`, `StatusCode`, `MethodDescriptor` and `MethodType` are exported by `grpc-web`.

**Suggested fix**:
Either change the `index.d.ts` to reflect that some classes are not exported _or_ change `exports.js` to also export the classes that are exported in `index.d.ts`. I prefer the latter as that makes `grpc-web` more extendable.

If you think this is a good idea, I can create a merge-request for it in the coming week.

tinrab:
Added exports for classes with constructors so that they can be used with Typescript.

Fixes #1158.

### DIFF
--- a/packages/grpc-web/exports.js
+++ b/packages/grpc-web/exports.js
@@ -6,12 +6,16 @@
  */
 goog.module('grpc.web.Exports');
 
-const GrpcWebClientBase = goog.require('grpc.web.GrpcWebClientBase');
-const StatusCode = goog.require('grpc.web.StatusCode');
+const CallOptions = goog.require('grpc.web.CallOptions');
 const MethodDescriptor = goog.require('grpc.web.MethodDescriptor');
+const GrpcWebClientBase = goog.require('grpc.web.GrpcWebClientBase');
+const RpcError = goog.require('grpc.web.RpcError');
+const StatusCode = goog.require('grpc.web.StatusCode');
 const MethodType = goog.require('grpc.web.MethodType');
 
-module['exports']['GrpcWebClientBase'] = GrpcWebClientBase;
-module['exports']['StatusCode'] = StatusCode;
+module['exports']['CallOptions'] = CallOptions;
 module['exports']['MethodDescriptor'] = MethodDescriptor;
+module['exports']['GrpcWebClientBase'] = GrpcWebClientBase;
+module['exports']['RpcError'] = RpcError;
+module['exports']['StatusCode'] = StatusCode;
 module['exports']['MethodType'] = MethodType;


packages/grpc-web/exports.js
```diff
@@ -6,12 +6,16 @@
  */
 goog.module('grpc.web.Exports');
 
+const CallOptions = goog.require('grpc.web.CallOptions');
+const MethodDescriptor = goog.require('grpc.web.MethodDescriptor');
 const GrpcWebClientBase = goog.require('grpc.web.GrpcWebClientBase');
+const RpcError = goog.require('grpc.web.RpcError');
 const StatusCode = goog.require('grpc.web.StatusCode');
-const MethodDescriptor = goog.require('grpc.web.MethodDescriptor');
 const MethodType = goog.require('grpc.web.MethodType');
```
sampajano, line 14: nit: do you mind help sort this line as well? :)

sampajano (approved):
Thanks so much for the change! 😃 (I'll wait for another important bug fix and create a bug fix release soon 😃)

Also adding @stanley-cheung to take a look too 😃

packages/grpc-web/exports.js
tinrab: The order reflects exports from `index.d.ts`? Is this ok?

sampajano: Ah thanks for the considerations :)

Personally i still prefer the exports to be alphabetically ordered in this file, given the connection to `index.d.ts` is an obscure one and would be harder to maintain (especially if we don't even document it here).

But i'm fine if you want to leave this as is in this PR :)

stanley-cheung (approved):

PR #1166, opened 2021-11-11
Add missing exports from RpcError and add test.

sampajano:
This complements https://github.com/grpc/grpc-web/pull/1164 to ensure the properties of `RpcError` are exported as well, and add tests.

(btw even without the `@export` annotation these 2 fields are exported today, but that's somewhat by accident -- it's due to them being referred by other classed rather than guaranteed.)


stanley-cheung (approved):
LGTM, very subtle!

Issue #1162, opened 2021-11-07
Inconsistent naming method with protobuf entities' "*_pb.d.ts"


xinxiao:
Should we consider changing the naming convention here to make things more consistent?

https://github.com/grpc/grpc-web/blob/5831a96b5665e17516c11ff44c0e13947bb92cad/javascript/net/grpc/web/generator/grpc_generator.cc#L1482

sampajano:
I kinda agree.. :)

Do you mind send a PR? Thanks :)

sampajano:
Once that's done, we can probably also clean up the comments here:

https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/net/grpc/gateway/examples/echo/ts-example/client.ts#L22-L27

And have our integration tests test both CommonJS+dts and Typescript import styles.

xinxiao:
#1168 PR created. 

I found it a bit hard to test the change. I tried to run the series of example docker containers to see if the change is valid but,

a. Couldn't find a docker test scenario for `--grpc-web_out=import_style=typescript` so I modified the case locally for `commonjs+dts` to fit.
b. The `prereqs` image only works for x86 machines but sadly what I have is an arm mac :( I wonder if we could use something like [bazelisk](https://github.com/bazelbuild/bazelisk) to wrap a good bazel installment for the testing environment?

sampajano:
> #1168 PR created.

Thanks so much for the PR! 😃

> I found it a bit hard to test the change. I tried to run the series of example docker containers to see if the change is valid but,
> 
> a. Couldn't find a docker test scenario for `--grpc-web_out=import_style=typescript` so I modified the case locally for `commonjs+dts` to fit.

Thanks for doing the local tests.. Indeed we don't have integration tests for `import_style=typescript`..

We do have some unit tests like the [eval_test](https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/packages/grpc-web/test/eval_test.js#L60) and [tsc_test](https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/packages/grpc-web/test/tsc_test.js#L103) that you'd need to update.

> b. The `prereqs` image only works for x86 machines but sadly what I have is an arm mac :( I wonder if we could use something like [bazelisk](https://github.com/bazelbuild/bazelisk) to wrap a good bazel installment for the testing environment?

[...]

PR #1168, opened 2021-11-13
Use consistent names for generated stubs in typescript.

xinxiao:


net/grpc/gateway/examples/echo/ts-example/client.ts
```diff
@@ -19,13 +19,8 @@
 import * as grpcWeb from 'grpc-web';
 import * as $ from 'jquery';
 
-// Uncomment either one of the following:
-// Option 1: import_style=commonjs+dts
 import {EchoServiceClient} from './echo_grpc_web_pb';
 
-// Option 2: import_style=typescript
```
sampajano, line 26: Thanks for cleaning up!

To test this locally, you could modify the ts-client [Dockerfile](https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/net/grpc/gateway/docker/ts_client/Dockerfile#L23) and manually run the echo server using `docker-compose up --build node-server envoy ts-client`, and see if the example still works (visiting `http://localhost:8081/echotest.html`).

We should probably add integration for the TS import style too soon :)

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -1478,8 +1470,7 @@ string GeneratorOptions::OutputFile(const string& proto_file) const {
     string directory;
     string basename;
     PathSplit(proto_file, &directory, &basename);
-    return directory + UppercaseFirstLetter(StripProto(basename)) +
-           "ServiceClientPb.ts";
+    return directory + StripProto(basename) + "_grpc_web_pb.ts";
```
sampajano, line 1473: We have some unit tests like the [eval_test](https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/packages/grpc-web/test/eval_test.js#L60) and [tsc_test](https://github.com/grpc/grpc-web/blob/1efe741695e7d330a4a56fe2e223fc027f83705c/packages/grpc-web/test/tsc_test.js#L103) that you'd probably need to update.

I wonder if you're able to run the [`run_basic_test.sh`](https://github.com/grpc/grpc-web/blob/master/scripts/run_basic_tests.sh) script locally for testing?

sampajano (commented):
Thanks so much for trying to unify the filename style! :)

@stanley-cheung Do you see any potential issue with this change?


@stanley-cheung I want to double check with you if there's any potential side effect with this change?

stanley-cheung:
Sorry for not reviewing this earlier. I thought I might disagree on this one but I could be wrong (or have outdated/incorrect info in the past).

I thought that it's more common for TS project to have filenames more like `EchoServiceClientPb.ts` rather than the commonJS underscore style.

Also, this might not be as big as a consistency issue in that - most projects using this will pick one style (commonJS vs TS) over another and won't mix. So if they all use `--import_style=typescript`, then they are consistent within that. I don't see anyone should try to mix `--import_style=typescript` with `--import_style=commonjs+dts` within the same project.

So unless TS projects actually use the understore style filenames, I'd rather keep the status quo.

sampajano:
Thanks for catching the TS style convention, Stanley. :)

@xinxiao sorry that i didn't catch this in the discussion in https://github.com/grpc/grpc-web/issues/1162. If Stanley's comment makes sense to you then maybe we can drop this..

(One alternative discussed is to convert both to `EchoServiceClientPb` style but then it creates inconsistency with commonjs import style and also a breaking change for users to migrate so it doesn't seem ideal either.)

xinxiao:
Makes sense; I'll close this PR.

PR #1170, opened 2021-11-18
Pin typescript @`4.4.4` to fix CI

sampajano:
CI broke today due to [`typescript`](https://www.npmjs.com/package/typescript) updating to `4.5.2` ([release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/)). Pinning it down to `4.4.4` to fix CI.

It appears that `4.5.2` stop generating `echo_grpc_web_pb.js` and `echo_pb.js` files in the `dist/` directory, despite there being corresponding `.d.ts` files.

Ideas on why and a proper fix (e.g. maybe we should just copy the files into `dist/` ourselves?) are very welcome.. :)

PR #1171, opened 2021-11-18
Use latest typescript and manually move `*_pb.js` files into `dist/`

sampajano:
See https://github.com/grpc/grpc-web/pull/1170 for the issue caused by change in typescript `4.5.2` and why this is needed.


stanley-cheung (approved):
LGTM

Issue #1153, opened 2021-10-22
Problem with deadline after switching to 1.3.0


kolesnikov-sergey:
After upgrading to version 1.3.0, when specifying a deadline, the request is not sent to the server, the promise is not completed. It works correctly without specifying a deadline

Example to reproduce https://github.com/kolesnikov-sergey/deadline-example/blob/master/src/index.js

bram-abe:
> Thanks for the report and sorry for the late reply!
> 
> I'm able to reproduce this issue and will work on a fix soon! 😃

Thank you

yoitsro:
Any news on this? 🙏

sampajano:
> Any news on this? 🙏

Thanks for the interest and sorry for the delay..

After digging deeper it turns out that this is caused by an issue deep down in the Closure library (with `goog.global`) and how it's incompatible with how we're packaging and importing the grpc web library..

While that issue will take a while to fix, i'll make a short-term fix and do a bug fix release soon.

Thanks again for the patience!

bram-abe:
> > Any news on this? 🙏
> 
> Thanks for the interest and sorry for the delay..
> 
> After digging deeper it turns out that this is caused by an issue deep down in the Closure library (with [`goog.global`](https://github.com/google/closure-library/blob/b5188f095a75338eeb09ab899fc94946632f453c/closure/goog/base.js#L47-L56)) and how it's not currently compatible with how we're packaging and importing the grpc-web library..
> 
> While that issue will take a while to fix, i'll make a short-term fix and do a bug fix release soon.
> 
> Thanks again for the patience!

thank you for update, will be waiting for fix release soon.

sampajano:
This issue should've been fixed by the above PR.

I've published a RC release [`1.3.1-rc`](https://www.npmjs.com/package/grpc-web/v/1.3.1-rc) (unfortunately it's tagged as `latest` too 🥲) which contains the fix. Please lemme know if it works (or not).

I'll try to cut a official 1.3.1 release soon too.. :)

Thanks again! :)

dalemarchand:
> This issue should've been fixed by the above PR.
> 
> I've published a RC release [`1.3.1-rc`](https://www.npmjs.com/package/grpc-web/v/1.3.1-rc) which contains the fix. Please try it out and lemme know if it works (or not). 😃
> 
> I'll try to cut a official 1.3.1 release soon too :)
> 
> Thanks again for the patience! :)

[...]

PR #1172, opened 2021-11-19
Fix issue where no RPC is issued when `deadline` is specified.

sampajano:
### The issue

This is a temporary hack to fix https://github.com/grpc/grpc-web/issues/1153, which is caused by [`goog.global`](https://github.com/google/closure-library/blob/f7c1a98dfed9e77a458817ff877c7336079ce3bf/closure/goog/base.js#L47-L56) not properly pointing to the proper global scope (e.g. `window`) when grpc-web is being imported as a CommonJS module (because [`this`](https://github.com/google/closure-library/blob/f7c1a98dfed9e77a458817ff877c7336079ce3bf/closure/goog/base.js#L48-L53) points to the module scope instead of actual global scope (i.e. `globalThis`)).

`goog.Timer.defaultTimerObject` is set to [point to](https://github.com/google/closure-library/blob/f7c1a98dfed9e77a458817ff877c7336079ce3bf/closure/goog/timer/timer.js#L112) `goog.global`, and that's why it throws when `setTimeout` is [called](https://github.com/google/closure-library/blob/f7c1a98dfed9e77a458817ff877c7336079ce3bf/closure/goog/timer/timer.js#L290) on it.

### The fix

* In this fix, we'll try to use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) when possible, and if it's not available (pre-2020 browsers, etc.), fallback to [`self`](https://developer.mozilla.org/en-US/docs/Web/API/Window/self), which should work in both browsers and workers.
* Also added an interop test to verify the fix works (the new test fails without the fix)

### DIFF
--- a/packages/grpc-web/exports.js
+++ b/packages/grpc-web/exports.js
@@ -19,3 +19,9 @@ module['exports']['GrpcWebClientBase'] = GrpcWebClientBase;
 module['exports']['RpcError'] = RpcError;
 module['exports']['StatusCode'] = StatusCode;
 module['exports']['MethodType'] = MethodType;
+
+// Temporary hack to fix https://github.com/grpc/grpc-web/issues/1153, which is
+// caused by `goog.global` not pointing to the global scope when grpc-web is
+// being imported as a CommonJS module.
+// TODO: Remove this hack after `goog.global` is fixed.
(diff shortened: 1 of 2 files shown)


stanley-cheung (approved):
LGTM

PR #1173, opened 2021-11-19
Update envoy version to 1.20

stanley-cheung:
Sometimes Envoy has breaking changes when upgrading between versions. Looks like we are clear this time.


sampajano (approved):
Thanks :)

PR #1184, opened 2022-01-10
Correctly support proto3 optional fields in commonjs+dts .d.ts output

### LINKED ISSUE #1072: .d.ts typescript definition doesn't match js with proto3 optional fields
Given a proto file like

```protobuf
syntax = "proto3";

message MyMessage {
  optional float my_prop = 1;
}
```

Running the generator with options `--js_out=import_style=commonjs,binary:. --grpc-web_out=import_style=commonjs+dts,mode=grpcwebtext:.` the generated `_pb.js` file correctly has the `get`, `set`, `has`, and `clear` methods for interacting with the property. The `_pb.d.ts` file, however, describes the property using the fallback enum representation with `get`, `set`, and a `MyPropCase` enum for describing the `_NOT_SET` and set semantics.

While I understand that this is technically correct for a language generator to describe the proto message in this way (that's how backwards compatibility is supported), this doesn't really work in this case as the `.d.ts` file is supposed to describe the `.js` file.

Anyway, sorry if this is reported elsewhere, I couldn't find anything when searching.

mattnathan:
Fixes #1072

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -825,8 +825,9 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
                      "set$js_field_name$(value?: $js_field_type$): "
                      "$class_name$;\n");
     }
-    if ((field->type() == FieldDescriptor::TYPE_MESSAGE || field->has_optional_keyword()) &&
-        !field->is_repeated() && !field->is_map()) {
+    if (field->has_optional_keyword() ||
+        (field->type() == FieldDescriptor::TYPE_MESSAGE &&
+            !field->is_repeated() && !field->is_map())) {
       printer->Print(vars, "has$js_field_name$(): boolean;\n");
     }
     if (field->type() == FieldDescriptor::TYPE_MESSAGE || field->has_optional_keyword() ||
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -825,11 +825,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
                      "set$js_field_name$(value?: $js_field_type$): "
                      "$class_name$;\n");
     }
-    if (field->type() == FieldDescriptor::TYPE_MESSAGE &&
+    if ((field->type() == FieldDescriptor::TYPE_MESSAGE || field->is_optional()) &&
         !field->is_repeated() && !field->is_map()) {
       printer->Print(vars, "has$js_field_name$(): boolean;\n");
     }
-    if (field->type() == FieldDescriptor::TYPE_MESSAGE ||
+    if (field->type() == FieldDescriptor::TYPE_MESSAGE || field->is_optional() ||
         field->is_repeated() || field->is_map()) {
       printer->Print(vars, "clear$js_field_name$(): $class_name$;\n");
     }
@@ -852,10 +852,12 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
 
   for (int i = 0; i < desc->oneof_decl_count(); i++) {
     const OneofDescriptor* oneof = desc->oneof_decl(i);
-    vars["js_oneof_name"] = ToUpperCamel(ParseLowerUnderscore(oneof->name()));
-    printer->Print(
-        vars, "get$js_oneof_name$Case(): $class_name$.$js_oneof_name$Case;\n");
-    printer->Print("\n");
+    if (!oneof->is_synthetic()) {
+      vars["js_oneof_name"] = ToUpperCamel(ParseLowerUnderscore(oneof->name()));
+      printer->Print(
+          vars, "get$js_oneof_name$Case(): $class_name$.$js_oneof_name$Case;\n");
+      printer->Print("\n");
+    }
   }
 
   printer->Print(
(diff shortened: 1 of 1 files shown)


mattnathan:
I think this is correct, I'm not entirely sure how to build or test it. Guidance would be greatly appreciated.

mattnathan:
I've managed to build (via bazel) and test this locally against some proto files I have with optional fields and it correctly outputs hasXxx and clearXxx methods for those fields in the .d.ts files.

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -825,11 +825,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
                      "set$js_field_name$(value?: $js_field_type$): "
                      "$class_name$;\n");
     }
-    if (field->type() == FieldDescriptor::TYPE_MESSAGE &&
+    if ((field->type() == FieldDescriptor::TYPE_MESSAGE || field->has_optional_keyword()) &&
         !field->is_repeated() && !field->is_map()) {
```
sampajano, line 829: nit: i'd probably prefer to maybe rewrite the logic as follows:

```
(field->has_optional_keyword() ||
    (field->type() == FieldDescriptor::TYPE_MESSAGE && 
        !field->is_repeated() &&
        !field->is_map())
```

Probably doesn't make a huge difference, but it seems to make a little bit more sense to me.. :P

WDYT? :)

sampajano (approved):
@mattnathan Hi Matt! Thanks so much for working on this fix (and apologize for the delay)!! The change (logic) looks great to me! 😃

Lemme look a little bit into what are some good ways to unit test it. (If not i'd say just merge it anyways since it's probably low risk :D)

@stanley-cheung Lemme know if you have some ideas on how you would test this change too! Thanks! :)

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -852,10 +852,12 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
 
   for (int i = 0; i < desc->oneof_decl_count(); i++) {
     const OneofDescriptor* oneof = desc->oneof_decl(i);
-    vars["js_oneof_name"] = ToUpperCamel(ParseLowerUnderscore(oneof->name()));
-    printer->Print(
-        vars, "get$js_oneof_name$Case(): $class_name$.$js_oneof_name$Case;\n");
-    printer->Print("\n");
+    if (!oneof->is_synthetic()) {
```
sampajano, line 855: Btw nice job finding the `is_synthetic()` API!😃

javascript/net/grpc/web/generator/grpc_generator.cc
mattnathan: I'm not sure.

The semantics of the two conditions is only the same when you factor in the [proto3 language spec](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#presence-in-proto3-apis) disallowing `optional` on map or repeated fields.

I read it as _when the field is a message or optional, but not when repeated or a map_ as opposed to _when the field is optional (which implies it's not repeated or a map), or it's a message but not repeated or a map_.

I'll leave the final decision up to you though :)

javascript/net/grpc/web/generator/grpc_generator.cc
mattnathan: Thanks, you forget what it's like to find docs via Google instead of using your IDE 😂

javascript/net/grpc/web/generator/grpc_generator.cc
sampajano: Matt - just a reminder about this comment :)

javascript/net/grpc/web/generator/grpc_generator.cc
sampajano: Yes right.. 😄

javascript/net/grpc/web/generator/grpc_generator.cc
sampajano: > The semantics of the two conditions is only the same when you factor in the proto3 language spec disallowing optional on map or repeated fields.

Yeah i think that was my exactly concern.. -- that `optional` is mutually exclusive with `repeated` and `map` so i don't think they should be in a `AND` relationship.. :)

> I read it as when the field is a message or optional, but not when repeated or a map as opposed to when the field is optional (which implies it's not repeated or a map), or it's a message but not repeated or a map.

Yeah right.. although i felt that's a little long-winded.. (and i had some challenge understanding that too) 😅

Where i was thinking as: _when the field is optional, or a message that is not repeated or a map_.

I feel the latter is simpler and easier to reason. :)

If you are ok both ways, do you mind making the change? Thanks! :)

mattnathan: The change has been made

Not sure what the process is around resolving conversations, so I'll leave this thread here

mattnathan:
Sorry for any delays in responses, UK based over here

javascript/net/grpc/web/generator/grpc_generator.cc
sampajano: Thanks a lot for the change! I guess it doesn't hurt to just leave them open here :D

sampajano:
> Sorry for any delays in responses, UK based over here

Np at all! That's almost no delay at all (compared it taking 14 days for me to review :P)..

Thanks again for the contribution and being patient with the review! :)

(I will cut a release very soon (maybe today or latest next week) which contains this fix. Thanks! :))

PR #1191, opened 2022-01-28
Remove outdated limitation regarding format mode

trysetnull:
The README states that "only unary calls are supported for now" regarding the `mode=grpcweb` (binary protobuf format).
I have a server streaming call and am generating typescript code with the following, it seems to work.

```
protoc \
    --proto_path=./ \
    --plugin=protoc-gen-grpc-web=/usr/local/grpc-web/protoc-gen-grpc-web \
    --js_out=import_style=commonjs,binary:${OUT_DIR} \
    --grpc-web_out=import_style=typescript,mode=grpcweb:${OUT_DIR} \
    ${INPUT_FILE}
```

Also it seems this is tested: scripts/docker-run-interop-tests.sh

Am I missing something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,13 +111,11 @@ For more information about the gRPC-web wire format, see the
 
   - `Content-type: application/grpc-web-text`
   - Payload are base64-encoded.
-  - Both unary and server streaming calls are supported.
 
 `mode=grpcweb`: A binary protobuf format is also supported.
 
   - `Content-type: application/grpc-web+proto`
   - Payload are in the binary protobuf format.
-  - Only unary calls are supported for now.
 
 ## How It Works
 


sampajano:
Hi! thanks so much for your PR!

Unfortunately, server streaming should STILL NOT be working for the binary mode...

The reason is that XHR doesn't give incremental backs unless the response has a text encoding.

> I have a server streaming call and am generating typescript code with the following, it seems to work.

When server streaming is working, the responses should come in chunks. In binary mode, all of the response chunks will come back _all at once_ at the end of the stream. Is that maybe the behavior that you're seeing?

> Also it seems this is tested: scripts/docker-run-interop-tests.sh

Yes, binary mode is indeed tested in the interop-test! Although, server streaming was specifically disabled for binary mode testing ([here](https://github.com/grpc/grpc-web/blob/eb313c1f3d81ab3e7b36ffffa8a93cf689b049bb/test/interop/interop_client.js#L183-L184)).. 😃

(FYI I actually tried this and it was working for me for a while.. I was super confused and then it turned out that my browser was caching the `main.js` from a previous session.. 😂 I'll sent a PR to disable the caching so this shouldn't confuse anymore.. Just want to mention in case this was your problem too..)

PR #1203, opened 2022-02-09
Add Github Actions (workflows) for building grpc-web protoc plugins (Linux, MacOS, and Windows)

sampajano:
# Summary

For Linux, MacOS, and Windows:
* Added new Github Actions (Workflow) for building the `protoc-gen-grpc-web` plugin to be used for releases.
* Workflow can be manually triggered with a `version_number` param (part of binary name)
* sha256 checksum are also generated and verified (except Windows)
* Plugin binary and sha256 are uploaded for manual download

## Platform specifics

* Linux - plugin built using Docker
* MacOS - plugin built using Makefile
  * Protobuf installed using `make install` from git submodule
* Windows - plugin built using Bazel

### DIFF
--- a/.github/workflows/make-plugin-linux.yml
+++ b/.github/workflows/make-plugin-linux.yml
@@ -1,0 +1,32 @@
+name: Make Linux Plugin
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_number:
+        description: 'Version number'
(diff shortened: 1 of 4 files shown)


stanley-cheung (approved):
Nice work!! Thanks!

PR #1210, opened 2022-02-25
Make the static flag overridable

06kellyjac:
Forcing the `-static` flag breaks nixpkgs builds so it's important to be able to disable it


sampajano (approved):
thanks! :)

Issue #1159, opened 2021-11-01
Provide prebuilt binaries for ARM in GitHub Releases


hronro:
Since M1 chip becomes more and more popular, it would be great if we can download prebuilt binaries for ARM in GitHub Releases.

sampajano:
This is a very reasonable ask... We'll keep that in mind and see if we can do it soon (but I don't have a M1 chip laptop yet..) :)

sampajano:
In the meantime, you could just use `make client` on Mac from the project root (which in turn uses this [Makefile](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/generator/Makefile)) to build the plugin on MacOS.

(You might need to `brew install protobuf` first to install the protobuf lib.)

tiero:
> you could just use make client on Mac

In the Makefile you linked there is no `client` target AFAICT

sampajano:
> > you could just use make client on Mac
> 
> In the Makefile you linked there is no `client` target AFAICT

You're right. :)

`make client` from project root would first invoke this [Makefile](https://github.com/grpc/grpc-web/blob/master/Makefile), and will only in turn use the Makefile i linked above.

Sorry for the confusion :)

rafaelmagalhaes:
any updates on a PR for this fix?

sampajano:
> any updates on a PR for this fix?

This is not a code change so no PR is required. Our next step is to find a M1 chip MacBook to build the binary from :)

For now, please use the Makefile to build the plugin as mentioned [above](https://github.com/grpc/grpc-web/issues/1159#issuecomment-956690764).

Thanks :)

hronro:
@sampajano 
Cool.  And I wonder if Linux ARM binaries and Windows ARM binaries are going to be supported as well?

sampajano:
> @sampajano Cool. And I wonder if Linux ARM binaries and Windows ARM binaries are going to be supported as well?

Not sure since i don't have configurations (or know they exists 😂) .. Might have to rely on the community to help test it out once we have a ARM binary :)

tiero:
@sampajano  I have M1 Apple Silicon, I can build for you and sign&release eventually.

I git cloned the repository, but looks like I need to load some google protobuf somewhere? Didn't find instructions

```
❯ make plugin          
cd "/Users/tiero/projects/tiero/grpc-web"/javascript/net/grpc/web/generator && make
[...]

PR #1214, opened 2022-03-16
ci: add support for linux-arm64 build

strophy:
This PR (my first here!) contains GitHub Actions CI to build `linux-arm64` artifacts in CI, partially addressing some of the issues raised in #1159.

While writing this I was unable to use (deprecated) `docker-compose` or the newer `docker compose` or `docker bake` system to build a compose file with `depends_on`. I think this may be due to different behaviour between Docker classic builds and buildx, but buildx is required for multi-arch images. As a result, I had to explicitly build and push each Dockerfile manually and in sequence so running the action will now also push updated `prereq` and `protoc_plugin` images to Docker Hub. Together with running in an emulator, this is quite slow (although caching improves this on subsequent runs), so I'm open to suggestions for improvements. It might also make sense to separate out building docker images and building a release, depending on how often you release.

Several underlying versions and dependencies had to be bumped in order to build successfully on arm64, and I switched from `bazel` to `bazelisk`, which was also necessary to build.

### Dependency Updates:
- All node base images now use `node:16.15.0-bullseye`
- `prepare` stage now uses `buildpack-deps:bullseye` for build toolchain consistency, although I think this bump is not strictly necessary
- Buildifier now at `5.1.0` because `1.0.0` does not support arm64
- Bazel `4.1.0` replaced with Bazelisk `1.11.0` (installs latest Bazel, currently `5.2.0`)
- `google-closure-compiler` JS dependency bumped to `~20220301.0.0` to avoid build error
- `google-protobuf` JS dependency bumped to `3.19.4` to match existing dependency used in `prereqs` image


strophy:
I believe this PR can now be improved with the techniques described here: https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/

Before I do this, is this project still maintained? Is this feature even wanted or likely to be merged? Just checking before I do any more work, since I haven't seen any interaction here.

packages/grpc-web/package.json
```diff
@@ -27,10 +27,10 @@
...
     "@types/google-protobuf": "~3.7.0",
     "command-exists": "~1.2.8",
-    "google-closure-compiler": "~20200224.0.0",
-    "google-closure-deps": "~20210601.0.0",
-    "google-closure-library": "~20210808.0.0",
-    "google-protobuf": "~3.14.0",
+    "google-closure-compiler": "^20220301.0.0",
+    "google-closure-deps": "^20220202.0.0",
+    "google-closure-library": "^20220202.0.0",
+    "google-protobuf": "~3.19.4",
```
sampajano, line 33: I'm curious why some of the JS dependency has to be updated for building on arm64? Are they somehow relevant?

.github/workflows/make-plugin-linux.yml
```diff
@@ -13,18 +13,95 @@ jobs:
...
+      id: docker_meta_prereqs
+      uses: docker/metadata-action@v3
+      with:
+        images: grpcweb/prereqs
+        tags: type=semver,pattern={{version}},value=${{ github.event.inputs.version_number }}
+        flavor: latest=true
+
+    - name: Build and push Docker image prereqs
+      id: docker_build_prereqs
+      uses: docker/build-push-action@v2
```
sampajano, line 47: curious where are the images being pushed to?

net/grpc/gateway/docker/protoc_plugin/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM grpcweb/prereqs
+FROM grpcweb/prereqs:latest
```
sampajano, line 15: not sure if this change would work with local builds with `docker compose`..

sampajano:
@strophy Thanks so much for the contrib! Apologies on the delay -- i saw this PR a while ago but i was unfamiliar with many of the concepts / tools used here so i didn't get around to it until now. Sorry again!

Pardon my ignorance, but i wonder how popular is linux-arm64? What's its most common use cases?

In general, we're aiming to provide pre-built binaries for the most popular platforms, and so i'm not sure we're willing to pay for the complexity of maintaining for less popular ones, especially if the complexity is high.

There are a few things in this PR that i'm not too sure about.. and i wonder if they can be addressed somehow:

1. It's complicating the workflow by not using `docker compose` -- that might be able to be addressed if you completely create a new workflow rather than reusing the existing one (and make it an experimental one).

2. Using some seemingly less official packages (e.g. [tonistiigi/binfmt](https://github.com/tonistiigi/binfmt)) and requires temporary hacks like https://github.com/bazelbuild/bazel/issues/11379 -- again showing the support is maybe too cutting-edge.

3. Updating many dep versions (e.g. `buildpack-deps`, `node`, `buildifier`) which i'm not sure are all necessary and desired... (they maybe are, but again showing that by supporting linux-arm64 we're moving a lot closer to the cutting edge.. :))

Also, curious how is this PR related to https://github.com/grpc/grpc-web/issues/1159 as you've mentioned in the PR description? Do you think this will be a progressive step to getting us M1 support? (and if so how big is the gap there?)

In any case, thanks so much for the PR again!

---

Btw, just ran CI and it's failing early. Could you take a look at the failure? Thanks!

packages/grpc-web/package.json
strophy: Yes, previous versions resulted in build failure on arm64

net/grpc/gateway/docker/protoc_plugin/Dockerfile
strophy: You're right, this is a hack and it still doesn't work right, I'll leave a longer comment in the conversation flow

.github/workflows/make-plugin-linux.yml
strophy: This is also a workaround for the sequential build problem. The images will be pushed with the tags specified on line 52, i.e. `grpcweb/prereqs:<version>`.

strophy:
Thanks for taking a look at this! I agree with all of the points you raised, this PR currently has too many hacks to merge at the moment. The point you raise in your response (1) above is the key of the issue though - in my testing, the current `docker-compose` or `docker compose` based build system also does not work, so I'm curious about how you used this to successfully build in the past. I am experiencing the following problem:

1. Clone the project and run `docker compose build prereqs protoc-plugin`
2. We would expect that Docker Compose respects the `depends_on` statements in `docker-compose.yml`, but this is not the case. `depends_on` is only respected by the `up` and `stop` commands, it is ignored by the `build` command. [Documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on). [Discussion](https://github.com/docker/compose/issues/6332).
3. The result is that instead of waiting for `grpcweb/prereqs` to be built, it instead immediately pulls the latest version from Docker Hub and uses this to start the `grpcweb/protoc-plugin` build in parallel with `grpcweb/prereqs`. This causes my build to fail, because the online version of the image is incompatible with the changes my build requires.
4. Instead, if I "manually" target the steps in two commands like `docker compose build prereqs && docker compose build protoc-plugin` then the build is successful, because `grpcweb/prereqs` will exist in the locally available images when the second command runs.

This is why I separated the build CI into steps and stopped using Docker Compose. Can you clarify if this is expected behavior for you? Have you been able to get `docker compose build` to respect the `depends_on` statements somehow?

strophy:
Regarding the usefulness of the `linux/arm64` this will make it possible to run images depending on grpc-web binaries on M1 Macs in a Docker environment with emulation or Rosetta. It is also frequently used in AWS datacentres, where the Graviton 1, 2 and 3 family of CPUs is available for Linux workloads at a 40% price saving over x86. Finally, devices like the Raspberry Pi and other low-power ARM boards are increasingly prevalent.

.github/workflows/make-plugin-linux.yml
strophy: Pushing is now no longer required, the `protoc_plugin` build successfully waits for the `prereq` build to complete.

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -49,10 +51,11 @@ RUN ./scripts/init_submodules.sh
 ######################################
 # Stage 2: Copy source files and build
 ######################################
-FROM node:12.22.6-stretch AS copy-and-build
+FROM node:16-bullseye AS copy-and-build
```
strophy, line 54: This change prevents the following error under arm64 when running the closure compiler as follows:
```
#0 159.1 /github/grpc-web/packages/grpc-web/node_modules/.bin/google-closure-compiler --js=exports.js --js=../../javascript --js=node_modules/google-closure-library --entry_point=grpc.web.Exports --externs=externs.js --dependency_mode=PRUNE --compilation_level=ADVANCED_OPTIMIZATIONS --generate_exports --export_local_property_definitions --js_output_file=index.js
#0 179.0 npm ERR! code ENOENT
#0 179.0 npm ERR! syscall access
#0 179.0 npm ERR! path /github/grpc-web/packages/grpc-web/node_modules/grpc-web.js
#0 179.1 npm ERR! errno -2
#0 179.1 npm ERR! enoent ENOENT: no such file or directory, access '/github/grpc-web/packages/grpc-web/node_modules/grpc-web.js'
#0 179.1 npm ERR! enoent This is related to npm not being able to find a file.
#0 179.1 npm ERR! enoent 
```

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -61,12 +64,10 @@ COPY --from=prepare /usr/local/bin/protoc /usr/local/bin/
...
 COPY --from=prepare /github/grpc-web/third_party /github/grpc-web/third_party
 
-RUN wget -nv -O bazel-installer.sh \
-  https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/\
-bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
-  chmod +x ./bazel-installer.sh && \
-  ./bazel-installer.sh && \
-  rm ./bazel-installer.sh
+RUN wget -nv -O /usr/local/bin/bazelisk \
+  https://github.com/bazelbuild/bazelisk/releases/download/v$BAZELISK_VERSION/bazelisk-linux-$TARGETARCH && \
```
strophy, line 68: Use of bazelisk is required, since `bazel-$BAZEL_VERSION-installer-linux-$arch` does not exist for arm64

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -19,22 +19,24 @@
 # Stage 1: Fetch binaries
 ######################################
 # node:... Docker image is based on buildpack-deps:stretch
-FROM buildpack-deps:stretch AS prepare
+FROM buildpack-deps:bullseye AS prepare
```
strophy, line 22: This change was made to match base image versions with the requirement to use `node16:bullseye` below

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -19,22 +19,24 @@
...
 
-ARG BUILDIFIER_VERSION=1.0.0
+ARG BUILDIFIER_VERSION=5.1.0
```
strophy, line 24: This change is required because Buildifier 1.0.0 does not exist for arm64

packages/grpc-web/package.json
strophy: Here is the specific build error if using an older version of the closure compiler dependency under arm64:
```
#0 88.56 /github/grpc-web/packages/grpc-web/node_modules/.bin/google-closure-compiler --js=exports.js --js=../../javascript --js=node_modules/google-closure-library --entry_point=grpc.web.Exports --externs=externs.js --dependency_mode=PRUNE --compilation_level=ADVANCED_OPTIMIZATIONS --generate_exports --export_local_property_definitions --js_output_file=index.js
#0 90.60 Error: EISDIR: illegal operation on a directory, read
#0 90.60     at Object.readSync (fs.js:568:3)
#0 90.60     at tryReadSync (fs.js:353:20)
#0 90.60     at Object.readFileSync (fs.js:390:19)
#0 90.60     at _Sd (/github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler-js/jscomp.js:7552:192)
#0 90.60     at YSd (/github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler-js/jscomp.js:9180:449)
#0 90.60     at sS (/github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler-js/jscomp.js:2442:29)
#0 90.60     at vS (/github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler-js/jscomp.js:3940:44)
#0 90.60     at /github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler-js/jscomp.js:4181:46
#0 90.60     at CompilerJS.run (/github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler/lib/node/closure-compiler-js.js:54:17)
#0 90.60     at /github/grpc-web/packages/grpc-web/node_modules/google-closure-compiler/cli.js:164:31 {
#0 90.60   errno: -21,
#0 90.60   syscall: 'read',
#0 90.60   code: 'EISDIR'
#0 90.60 }
#0 109.6 npm ERR! code ENOENT
#0 109.6 npm ERR! syscall access
#0 109.6 npm ERR! path /github/grpc-web/packages/grpc-web/node_modules/grpc-web.js
#0 109.6 npm ERR! errno -2
#0 109.6 npm ERR! enoent ENOENT: no such file or directory, access '/github/grpc-web/packages/grpc-web/node_modules/grpc-web.js'
#0 109.6 npm ERR! enoent This is related to npm not being able to find a file.
#0 109.6 npm ERR! enoent 
```

strophy:
Hello, I have refactored the PR to use `docker bake` and a `.hcl` file instead of `docker compose` and a `docker-compose.yml` file, since I believe that the logic used to build in the past incorrectly assumes that Docker Compose will respect `depends_on` when building (it does not). The build logic now actually waits for the `prereqs` stage to complete, instead of pulling a version based on a previous build from Docker Hub.

I have documented the various build errors that occur if I do not bump the dependency versions. It might be possible to resolve some of these in another way, but I am tempted to just use newer versions that take non-x86 targets into account rather than debug dependencies.

I still occasionally face the following issues:
- Build randomly fails in CI with following error. This seems to point to an OOM problem, but I am unable to reliably reproduce it and memory use never exceeds 2GB when running on a Github Actions runner. I would be curious if you see this error as well.
```
#29 177.7 Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/root/.cache/bazel/_bazel_root/94ede7e4be7ee112c06d5210902f7b7d/server/jvm.out')
#29 177.7 
#29 ERROR: process "/bin/sh -c bazelisk build --jobs=2 javascript/net/grpc/web/generator:protoc-gen-grpc-web &&   cp $(bazelisk info bazel-genfiles)/javascript/net/grpc/web/generator/protoc-gen-grpc-web   /usr/local/bin/protoc-gen-grpc-web" did not complete successfully: exit code: 37
```
- Cache does not always hit, we should maybe consider including a cache mount
- Output binaries still need testing

packages/grpc-web/package.json
sampajano: Aha interesting! Thanks for the info!

What about `google-closure-library`? What that cause an error is left unchanged?

sampajano:
Hi thanks for the reply!

> Thanks for taking a look at this! I agree with all of the points you raised, this PR currently has too many hacks to merge at the moment. The point you raise in your response (1) above is the key of the issue though - in my testing, the current `docker-compose` or `docker compose` based build system also does not work, so I'm curious about how you used this to successfully build in the past. I am experiencing the following problem:
> 
> 1. Clone the project and run `docker compose build prereqs protoc-plugin`
> 2. We would expect that Docker Compose respects the `depends_on` statements in `docker-compose.yml`, but this is not the case. `depends_on` is only respected by the `up` and `stop` commands, it is ignored by the `build` command. [Documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on). [Discussion](https://github.com/docker/compose/issues/6332).

Oh aha i see!! Thanks for providing the context!

Actually, historically i've always been using `docker-compose` (instead of `docker compose`) for both `build` and `up`, and as i tried it just now, it seems that `docker compose` has a significant different behavior in that it no longer respect the `depends_on` statement as you've pointed out.

This is actually something an interesting point that you've brought up because i've noticed the mismatch between build-time v.s. runtime dependency before, but i didn't know if there's an alternative way of specifying build-time dependency other than using `depends_on`.. (guess it won't work now either 🙃)

> 3. The result is that instead of waiting for `grpcweb/prereqs` to be built, it instead immediately pulls the already existing old version from Docker Hub and uses this to start the `grpcweb/protoc-plugin` build in parallel with `grpcweb/prereqs`. This causes my build to fail, because the Docker Hub version of the image is incompatible with the changes my build requires.
> 4. Instead, if I "manually" target the steps in two commands like `docker compose build prereqs && docker compose build protoc-plugin` then the build is successful, because `grpcweb/prereqs` will exist in the locally available images when the second command runs.
> 
> This is why I separated the build CI into steps and stopped using Docker Compose. Can you clarify if this is expected behavior for you? Have you been able to get `docker compose build` to respect the `depends_on` statements somehow? If not, what do you think about using the following approach as a way forward: https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#using-a-result-of-one-target-as-a-base-image-in-another-target

I really appreciate you finding an alternative solution using the buildx here! Although, i'm a bit hesitant in doing this across the board (other than just in this workflow), because we have a bunch of documentations in various sites that we need to update too ([example](https://grpc.io/docs/platforms/web/quickstart/)) if we decide to do that, and i'm not yet convinced that docker buildx bake + HCL is broadly adopted enough for that purpose.. (maybe it is? :))

For that reason, may I ask you to "fork" the workflow into a separate `make-plugin-linux-arm64.yml` config? So that it remains a bit "experimental" until we eventually decide to converge? Thanks!

.github/workflows/make-plugin-linux.yml
sampajano: Aha i see thanks for addressing this! :)

net/grpc/gateway/docker/prereqs/Dockerfile
sampajano: Aha thanks for the context!

Do you mind replacing all `12.22.6` references in this repo to the same node version tho (there are [3 files](https://cs.github.com/grpc/grpc-web?q=12.22.6) afaict)?

Ideally to a more specific dot release to ensure our test [here](https://github.com/grpc/grpc-web/blob/1779661dded5b918cdd84a216f708b93ebf2384c/packages/grpc-web/docker/jsunit-test/Dockerfile#L3-L4) will always be running the same node version as the node server.

net/grpc/gateway/docker/prereqs/Dockerfile
sampajano: Thanks for the explanation!

May i ask you a favor to list out *all* the dependency updates you've made in this PR in a bullet-point summary? (Feel free to mention all other notable changes you made as well!)

release.hcl
```diff
@@ -0,0 +1,15 @@
+group "release" {
```
sampajano, line 1: do you mind keeping this file under .github/workflow for now since it's only been used in that context? Thanks!

sampajano (commented):
thanks so much for the updates! Just a few comments! 😃

Mainly, as mentioned in a previous comment, would you mind forking out the workflow changes into a separate file for now? And will consider merging as we adopt buildx + hcl more broadly. Thanks!!

packages/grpc-web/package.json
strophy: Good catch! I tested again and it seems only `google-closure-compiler` needs to be bumped for the build to succeed.

strophy:
Thanks for reviewing! I've made the requested changes, but any testing is incredibly frustrating due to random failures running Bazelisk. I can't reproduce it reliably, but it only occurs when running the build of a whole file, stopping the build just before the problematic step and then running the command directly inside the partially built image always works perfectly, so there is no easy way for me to read the contents of the error log. I think it is best if we pause here and wait for the release of Bazel 5.2.0 which should contain numerous fixes allowing us to remove some of the hacks in this PR.

I think it's also best if we stop working on caching solutions and do that in a separate PR, since we can now build arm64 in a separate experimental workflow, speed is no longer a priority. 

List of updated dependencies:
- All node base images now use `node:16.15.0-bullseye`
- `prepare` stage now uses `buildpack-deps:bullseye` for build toolchain consistency, although I think this bump is not strictly necessary
- Buildifier now at `5.1.0` because `1.0.0` does not support arm64
- Bazel `4.1.0` replaced with Bazelisk `1.11.0` (installs Bazel `5.1.1`, currently broken, wait for `5.2.0`)
- `google-closure-compiler` JS dependency bumped to `~20220301.0.0` to avoid build error
- `google-protobuf` JS dependency bumped to `3.19.4` to match existing dependency used in `prereqs` image

I'm unable to spend much more time working on this, and I don't know how your CI system works. From the log output I can see it is using the Docker classic builder to attempt to process a Dockerfile that requires buildx to build, so some changes will be necessary here.

Also interesting to note how building from a compose file will likely be handled in the future [here](https://github.com/docker/compose/issues/9477), tl;dr it will parse the compose yml file, generate .hcl files and delegate everything to `buildx bake` anyway.

sampajano:
> Thanks for reviewing! I've made the requested changes, but any testing is incredibly frustrating due to random failures running Bazelisk. I can't reproduce it reliably, but it only occurs when running the build of a whole file, stopping the build just before the problematic step and then running the command directly inside the partially built image always works perfectly, so there is no easy way for me to read the contents of the error log. I think it is best if we pause here and wait for the release of Bazel 5.2.0 which should contain some fixes relating to running in an emulator, allowing us to remove some of the hacks in this PR.

Oh ok i see!! Very sorry to hear about the frustrations! I'm perfectly fine to pause here and wait for the 5.2.0 release! 😃

> 
> I think it's also best if we stop working on caching solutions and do that in a separate PR, since we can now build arm64 in a separate experimental workflow, speed is no longer a priority.

Sounds great to me too! Speed is indeed not an issue here at all (since we only do this every x months really..) 😃

> 
> List of updated dependencies:
> 
> * All node base images now use `node:16.15.0-bullseye`
> * `prepare` stage now uses `buildpack-deps:bullseye` for build toolchain consistency, although I think this bump is not strictly necessary
> * Buildifier now at `5.1.0` because `1.0.0` does not support arm64
> * Bazel `4.1.0` replaced with Bazelisk `1.11.0` (installs Bazel `5.1.1`, currently broken, wait for `5.2.0`)
> * `google-closure-compiler` JS dependency bumped to `~20220301.0.0` to avoid build error
> * `google-protobuf` JS dependency bumped to `3.19.4` to match existing dependency used in `prereqs` image

Thanks so much for the list of deps! I'll edit the PR summary to add these 😃

> 
> I'm unable to spend much more time working on this, and I don't know how your CI system (kokoro?) works. From the log output I can see it is using the Docker classic builder to attempt to process a Dockerfile that requires buildx to build, so some changes will be necessary here.

Do you think the Bazel 5.2.0 will help resolve the issue here?

I think our kokoro CI essentially just runs the [basic](https://github.com/grpc/grpc-web/blob/master/scripts/run_basic_tests.sh) and [interop](https://github.com/grpc/grpc-web/blob/master/scripts/run_interop_tests.sh) scripts. Do they work locally for you?

> 
> Also interesting to note how building from a compose file will likely be handled in the future [here](https://github.com/docker/compose/issues/9477), tl;dr it will parse the compose yml file, generate .hcl files and delegate everything to `buildx bake` anyway.

Oh wow! That's awesome! Would probably get a great improvement when it happens!

strophy:
Hi @sampajano Bazel 5.2.0 was released with a lot of ARM fixes, but I am still seeing the same error (`Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/root/.cache/bazel/_bazel_root/94ede7e4be7ee112c06d5210902f7b7d/server/jvm.out')`) when running in an emulator. I see you guys already completed arm64 CI using Zig which comes with an impressive speedup, so I think it's probably best to give up on Bazel for building arm64 for now! You can merge this as an alternative approach, or feel free to close this PR :smiley:

sampajano:
> Hi @sampajano Bazel 5.2.0 was released with a lot of ARM fixes, but I am still seeing the same error (`Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/root/.cache/bazel/_bazel_root/94ede7e4be7ee112c06d5210902f7b7d/server/jvm.out')`) when running in an emulator. I see you guys already completed arm64 CI using Zig which comes with an impressive speedup, so I think it's probably best to give up on Bazel for building arm64 for now! You can merge this as an alternative approach, or feel free to close this PR 😃

@strophy Thanks so much for checking again! Indeed the Zig solution is already checked in and seems to be a more self-contained change overall (and seems easy to apply to all existing platforms too). 😃

Thank you so much for the contributions here and iterations again! 😃 I'll close this PR for now but i think whatever discussion/change we have here would help provide a great context should we ever need to reevaluate our build method! 😃

Issue #1220, opened 2022-04-19
Envoy filter not working anymore with envoy >=v1.22


Gerrit91:
With the existing configuration sample from this project and envoy version v1.22, I get the following error message when starting up the proxy:

```
Didn't find a registered implementation for 'envoy.filters.http.grpc_web' with type URL: ''
```

With v1.21 everything works just fine. Is there any way to resolve this problem?

shivavishaal66215:
I'm stuck with the same error too. This might seem silly but someone on StackOverflow fixed this error by changing their envoy file from `envoy.yml` to `envoy.yaml`. @Gerrit91

sampajano:
@shivavishaal66215 that's funny indeed haha.. does changing `yml` -> `yaml` fix the issue for you as well? 😃

shivavishaal66215:
@sampajano Unfortunately it didn't work for me.

tomk9:
Based on https://github.com/envoyproxy/envoy/blob/2586aa6271811ba5653e7da82dfb62c85e740152/examples/front-proxy/service-envoy.yaml#L26:L28
@Gerrit91 @shivavishaal66215 try to add types:

<table>
<tr>
<td> v1.21 </td> <td> v1.22 </td>
</tr>
<tr>
<td>

```yaml
http_filters:
  - name: envoy.filters.http.grpc_web
  - name: envoy.filters.http.cors
  - name: envoy.filters.http.router
```

</td>
<td>

```yaml
http_filters:
  - name: envoy.filters.http.grpc_web
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.filters.http.grpc_web.v3.GrpcWeb
  - name: envoy.filters.http.cors
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
  - name: envoy.filters.http.router
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
```

</td>
</tr>

sampajano:
@tomk9 Thanks so much for the super timely PR and fix! Much appreciated!! 😃

Gerrit91:
Perfect! Thanks everyone for helping out with this one. :)

shivavishaal66215:
Thanks a lot folks. Amazing response time. 😄

tomk9:
[...]

PR #1222, opened 2022-04-21
Update envoy version to 1.22 (with config updates)

tomk9:
#1220

### DIFF
--- a/net/grpc/gateway/examples/echo/envoy.yaml
+++ b/net/grpc/gateway/examples/echo/envoy.yaml
@@ -35,9 +35,15 @@ static_resources:
                 max_age: "1728000"
                 expose_headers: custom-header-1,grpc-status,grpc-message
           http_filters:
-          - name: envoy.filters.http.grpc_web
-          - name: envoy.filters.http.cors
-          - name: envoy.filters.http.router
+            - name: envoy.filters.http.grpc_web
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.filters.http.grpc_web.v3.GrpcWeb
+            - name: envoy.filters.http.cors
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
+            - name: envoy.filters.http.router
+              typed_config:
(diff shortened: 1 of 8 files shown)


sampajano:
@tomk9  Thanks so much for the change!!

Interestingly, didn't find anything mentioning about a "breaking change" in 1.22 change log.. But it did mention a few things about "typed configuration" and "filter".. :)

https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22.0

net/grpc/gateway/examples/echo/envoy.yaml
```diff
@@ -35,9 +35,15 @@ static_resources:
                 max_age: "1728000"
                 expose_headers: custom-header-1,grpc-status,grpc-message
           http_filters:
-          - name: envoy.filters.http.grpc_web
-          - name: envoy.filters.http.cors
-          - name: envoy.filters.http.router
+            - name: envoy.filters.http.grpc_web
```
sampajano, line 38: just curious - does the name still matter now?

tomk9: I've checked. Looks like the name doesn't matter anymore.

net/grpc/gateway/examples/echo/envoy.yaml
```diff
@@ -35,9 +35,15 @@ static_resources:
...
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.filters.http.grpc_web.v3.GrpcWeb
```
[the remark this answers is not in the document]
tomk9: I was following the example
https://github.com/envoyproxy/envoy/blob/6931f18c3bacb0e41a64d2107b2c0d0cd683ad80/examples/front-proxy/service-envoy.yaml#L28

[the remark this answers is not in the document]
sampajano: Aha ok! Thanks for explaining! :)

net/grpc/gateway/examples/echo/envoy.yaml
sampajano: Oh i see! Good to know.. Thanks for checking! :)

PR #1223, opened 2022-04-21
Fix Envoy configs formatting.

sampajano:
Cleaning up the formatting of our Envoy configs following https://github.com/grpc/grpc-web/pull/1222.

- Also updated `run_interop_tests.sh` to clean up lingering docker containers to be more robust when ran locally.


dennycd (approved):
lgtm

dennycd (approved):

PR #1025, opened 2021-01-22
Allow null response

pro-wh:
We're using a custom codec, where null or undefined responses can be represented, and we use that for void methods, similar to how protobuf uses `Empty`.

This change allows grpc-web to accept responses that decode to `null` or `undefined`.

PR #1230, opened 2022-04-27
GrpcWebClientReadableStream: keep falsy data

pro-wh:
cross reference https://github.com/grpc/grpc-web/pull/1025

When using a different codec where primitive values are allowed (we use CBOR), the current implementation omits calling the `data` listeners on falsy values, such as `0` and `null`. But we'd like to have them.

Is there anything else that relies on the current behavior that omits these values?

---

Discussion of changes in `GrpcWebClientBase`:

Current implementation uses `setCallback_` in a special "useUnaryResponse" mode. In this mode, additional calls are made to the callback with `status` and `metadata` parameters, then a final time with both `error` and `response` set to `null`.

In order to support falsy and even null message, this changes the final callback call to pass `true` to a new explicit `trigger` parameter. `callback(null, null)` now means "no error, and the response was null" (and similar for other falsy responses).

`setCallback_` is private, and the only public wrapper does not allow the "useUnaryResponse" mode, so it should be safe to make this change.

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientbase.js
+++ b/javascript/net/grpc/web/grpcwebclientbase.js
@@ -123,18 +123,18 @@ class GrpcWebClientBase {
       let unaryStatus;
       let unaryMsg;
       GrpcWebClientBase.setCallback_(
-          stream, (error, response, status, metadata, unaryEndOfStream) => {
+          stream, (error, response, status, metadata, unaryResponseReceived) => {
             if (error) {
               reject(error);
+            } else if (unaryResponseReceived) {
+              unaryMsg = response;
             } else if (status) {
               unaryStatus = status;
             } else if (metadata) {
               unaryMetadata = metadata;
-            } else if (unaryEndOfStream) {
+            } else {
               resolve(request.getMethodDescriptor().createUnaryResponse(
                   unaryMsg, unaryMetadata, unaryStatus));
-            } else {
-              unaryMsg = response;
             }
           }, true);
     });
@@ -224,7 +224,14 @@ class GrpcWebClientBase {
    * @param {!ClientReadableStream<RESPONSE>} stream
    * @param {function(?RpcError, ?RESPONSE, ?Status=, ?Object<string, string>=, ?boolean)|
    *     function(?RpcError,?RESPONSE)} callback
-   * @param {boolean} useUnaryResponse
+   * @param {boolean} useUnaryResponse Pass true to have the client make
+   * multiple calls to the callback, using (error, response, status,
+   * metadata, unaryResponseReceived) arguments. One of error, status,
+   * metadata, or unaryResponseReceived will be truthy to indicate which piece
+   * of information the client is providing in that call. After the stream
+   * ends, it will call the callback an additional time with all falsy
+   * arguments. Pass false to have the client make one call to the callback
+   * using (error, response) arguments.
    */
   static setCallback_(stream, callback, useUnaryResponse) {
     let isResponseReceived = false;
@@ -272,11 +279,11 @@ class GrpcWebClientBase {
             message: 'Incomplete response',
           });
         } else {
-          callback(null, responseReceived);
+          callback(null, responseReceived, null, null, /* unaryResponseReceived= */ true);
         }
       }
       if (useUnaryResponse) {
-        callback(null, null, null, null, /* unaryEndOfStream= */ true);
+        callback(null, null);
       }
     });
   }
--- a/javascript/net/grpc/web/grpcwebclientbase.js
+++ b/javascript/net/grpc/web/grpcwebclientbase.js
@@ -123,18 +123,18 @@ class GrpcWebClientBase {
       let unaryStatus;
       let unaryMsg;
       GrpcWebClientBase.setCallback_(
-          stream, (error, response, status, metadata) => {
+          stream, (error, response, status, metadata, unaryEndOfStream) => {
             if (error) {
               reject(error);
-            } else if (response) {
-              unaryMsg = response;
             } else if (status) {
               unaryStatus = status;
             } else if (metadata) {
               unaryMetadata = metadata;
-            } else {
+            } else if (unaryEndOfStream) {
               resolve(request.getMethodDescriptor().createUnaryResponse(
                   unaryMsg, unaryMetadata, unaryStatus));
+            } else {
+              unaryMsg = response;
             }
           }, true);
     });
@@ -222,7 +222,7 @@ class GrpcWebClientBase {
    * @static
    * @template RESPONSE
    * @param {!ClientReadableStream<RESPONSE>} stream
-   * @param {function(?RpcError, ?RESPONSE, ?Status=, ?Object<string, string>=)|
+   * @param {function(?RpcError, ?RESPONSE, ?Status=, ?Object<string, string>=, ?boolean)|
    *     function(?RpcError,?RESPONSE)} callback
    * @param {boolean} useUnaryResponse
    */
@@ -276,7 +276,7 @@ class GrpcWebClientBase {
         }
       }
       if (useUnaryResponse) {
-        callback(null, null);  // trigger unaryResponse
+        callback(null, null, null, null, /* unaryEndOfStream= */ true);
       }
     });
   }
--- a/javascript/net/grpc/web/grpcwebclientbase_test.js
+++ b/javascript/net/grpc/web/grpcwebclientbase_test.js
@@ -102,6 +102,28 @@ testSuite({
     assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
   },
 
+  async testRpcResponseThenableCall() {
+    const xhr = new XhrIo();
+    const client = new GrpcWebClientBase(/* options= */ {}, xhr);
+    const methodDescriptor = createMethodDescriptor((bytes) => {
+      assertElementsEquals(DEFAULT_RPC_RESPONSE_DATA, [].slice.call(bytes));
+      return new MockReply('value');
+    });
+
+    const responsePromise = client.thenableCall(
+      'url', new MockRequest(), /* metadata= */ {}, methodDescriptor);
+    xhr.simulatePartialResponse(
+        googCrypt.encodeByteArray(new Uint8Array(DEFAULT_RPC_RESPONSE)),
+        DEFAULT_RESPONSE_HEADERS);
+    xhr.simulateReadyStateChange(ReadyState.COMPLETE);
+    const response = await responsePromise;
+
+    assertEquals('value', response.data);
+    const headers = /** @type {!Object} */ (xhr.getLastRequestHeaders());
+    assertElementsEquals(DEFAULT_UNARY_HEADERS, Object.keys(headers));
+    assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
+  },
+
   async testRpcFalsyResponseThenableCall_ForNonProtobufDescriptor() {
     const xhr = new XhrIo();
     const client = new GrpcWebClientBase(/* options= */ {}, xhr);
--- a/javascript/net/grpc/web/grpcwebclientbase_test.js
+++ b/javascript/net/grpc/web/grpcwebclientbase_test.js
@@ -102,6 +102,28 @@ testSuite({
     assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
   },
 
+  async testRpcFalsyResponseThenableCall_ForNonProtobufDescriptor() {
+    const xhr = new XhrIo();
+    const client = new GrpcWebClientBase(/* options= */ {}, xhr);
+    const methodDescriptor = createMethodDescriptor((bytes) => {
+      assertElementsEquals(DEFAULT_RPC_RESPONSE_DATA, [].slice.call(bytes));
+      return 0;
+    });
+
+    const responsePromise = client.thenableCall(
+      'url', new MockRequest(), /* metadata= */ {}, methodDescriptor);
+    xhr.simulatePartialResponse(
+        googCrypt.encodeByteArray(new Uint8Array(DEFAULT_RPC_RESPONSE)),
+        DEFAULT_RESPONSE_HEADERS);
+    xhr.simulateReadyStateChange(ReadyState.COMPLETE);
+    const response = await responsePromise;
+
+    assertEquals(0, response);
+    const headers = /** @type {!Object} */ (xhr.getLastRequestHeaders());
+    assertElementsEquals(DEFAULT_UNARY_HEADERS, Object.keys(headers));
+    assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
+  },
+
   async testDeadline() {
     const xhr = new XhrIo();
     const client = new GrpcWebClientBase(/* options= */ {}, xhr);
--- a/javascript/net/grpc/web/grpcwebclientbase_test.js
+++ b/javascript/net/grpc/web/grpcwebclientbase_test.js
@@ -75,6 +75,33 @@ testSuite({
     assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
   },
 
+  async testRpcFalsyResponse_ForNonProtobufDescriptor() {
+    const xhr = new XhrIo();
+    const client = new GrpcWebClientBase(/* options= */ {}, xhr);
+    const methodDescriptor = createMethodDescriptor((bytes) => {
+      assertElementsEquals(DEFAULT_RPC_RESPONSE_DATA, [].slice.call(bytes));
+      return 0;
+    });
+
+    const response = await new Promise((resolve, reject) => {
+      client.rpcCall(
+          'url', new MockRequest(), /* metadata= */ {}, methodDescriptor,
+          (error, response) => {
+            assertNull(error);
+            resolve(response);
+          });
+      xhr.simulatePartialResponse(
+          googCrypt.encodeByteArray(new Uint8Array(DEFAULT_RPC_RESPONSE)),
+          DEFAULT_RESPONSE_HEADERS);
+      xhr.simulateReadyStateChange(ReadyState.COMPLETE);
+    });
+
+    assertEquals(0, response);
+    const headers = /** @type {!Object} */ (xhr.getLastRequestHeaders());
+    assertElementsEquals(DEFAULT_UNARY_HEADERS, Object.keys(headers));
+    assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
+  },
+
   async testDeadline() {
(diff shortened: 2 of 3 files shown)


pro-wh:
oh oops this should skip if deserializing failed. I'll add some logic to do that

pro-wh:
there are more changes needed in `thenableCall`

sampajano:
Hi! Thanks for the contribution and sorry for the delay!

2 quick questions:

1. Curious how are you using a custom codec? I'm still somewhat new and didn't know that's possible 😃
2. Would it be possible for you to showcase your new use case by adding a test?

thanks!

pro-wh:
Hi, thanks for helping out with this.

## How to use a custom codec:

sample code https://github.com/oasisprotocol/oasis-sdk/blob/9cfbb9246ba1797e475728a48d580b0eb8a352d3/client-sdk/ts-web/core/src/client.ts#L22

We haven't been using .proto files to define the methods, so we construct a bunch of our own `MethodDescriptor` objects. Thus all it took for us was to pass the https://github.com/grpc/grpc-web/blob/1.3.1/javascript/net/grpc/web/methoddescriptor.js#L33 `responseDeserializerFn`

## Adding a test

Ya let me try adding one

sampajano:
> ## How to use a custom codec:
> sample code https://github.com/oasisprotocol/oasis-sdk/blob/9cfbb9246ba1797e475728a48d580b0eb8a352d3/client-sdk/ts-web/core/src/client.ts#L22
> 
> We haven't been using .proto files to define the methods, so we construct a bunch of our own `MethodDescriptor` objects. Thus all it took for us was to pass the https://github.com/grpc/grpc-web/blob/1.3.1/javascript/net/grpc/web/methoddescriptor.js#L33 `responseDeserializerFn`

Oh aha! Intersting! Thanks for the info! I guess that's one way of using the library (but without the codegen) 😃

I wonder if we mention the use of library in this way on any documentation? Or this is just one creative way you found out about using it? (I'm guess in the latter case I'm not sure if such use will be "officially supported" going forward, even if we could probably fix a few non-harming edge-cases right now)  😃

> ## Adding a test
> Ya let me try adding one

thanks!!

pro-wh:
- https://github.com/grpc/grpc-web/blob/1.3.1/packages/grpc-web/index.d.ts#L77 it's publicly listed in this manually written type declaration
- https://github.com/grpc/grpc-go/blob/master/Documentation/encoding.md#codecs-serialization-and-deserialization custom codecs are a normal part of grpc

sampajano:
> * https://github.com/grpc/grpc-web/blob/1.3.1/packages/grpc-web/index.d.ts#L77 it's publicly listed in this manually written type declaration
> * https://github.com/grpc/grpc-go/blob/master/Documentation/encoding.md#codecs-serialization-and-deserialization custom codecs are a normal part of grpc

Ah thanks for the info!

I did a quick search and found that the majority of mentioning for custom codec is for golang. (there are some mentions on other languages but nothing seems officially documented.) So personally i'd still consider this a borderline edge use case :) Happy to take small patches if it's easy to maintain, but probably no commitment for major feature support going forward :)

pro-wh:
aaah I just saw the new release and realized this was still pending

> So personally i'd still consider this a borderline edge use case :) Happy to take small patches if it's easy to maintain, but probably no commitment for major feature support going forward :)

we're happy with that too. if you could please review :pray:

sampajano:
> aaah I just saw the new release and realized this was still pending

aha right i've forgot as well :)

> > So personally i'd still consider this a borderline edge use case :) Happy to take small patches if it's easy to maintain, but probably no commitment for major feature support going forward :)
> 
> we're happy with that too. if you could please review 🙏

thanks for your interest here! On a 2nd thought, I'll need to check with the team on our stance on custom codec before getting back at you. Thanks!

pro-wh:
Sounds good, thanks for checking with the team

sampajano:
Of course!

Oh and actually, since you'll be depending on `MethodDescriptor`, i guess this will be related to https://github.com/grpc/grpc-web/issues/1285 -- where we're no longer exposing its API due to internal code size issue ([PR here](https://github.com/grpc/grpc-web/pull/1199)).

But we could explore our options there too..

In the meanwhile, maybe you could rebase your code onto 1.4.0 and see if it breaks you?

Thanks!

pro-wh:
interesting, thanks for the heads up. let me experiment with the changes in 1.4.0. if it's only from the typescript headers, we should be fine

pro-wh:
update: yes, it does break our `.getName` callsite. we will be switching back to `.name` when we upgrade to 1.4.0

sampajano:
> update: yes, it does break our `.getName` callsite. we will be switching back to `.name` when we upgrade to 1.4.0

aha interesting thanks for confirming! So i guess that would work but is somewhat a "hack".. :)

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -123,18 +123,18 @@ class GrpcWebClientBase {
...
             } else if (status) {
               unaryStatus = status;
             } else if (metadata) {
               unaryMetadata = metadata;
-            } else {
+            } else if (trigger) {
               resolve(request.getMethodDescriptor().createUnaryResponse(
                   unaryMsg, unaryMetadata, unaryStatus));
+            } else {
+              unaryMsg = response;
```
sampajano, line 137: I'm not sure we should set the `unaryMsg` on an `else {}` case — it seems a bit unsafe to set it before checking nullability (even though it might be ok based on existing callers).

Actually.. I'm curious what if you don't make this change? With the existing code, wouldn't the `stream.on('end' ...)` below eventually trigger a `callback(null, null)` call which will fall into this `else {}`clause, which will trigger the `resolve(...)` call?

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -123,18 +123,18 @@ class GrpcWebClientBase {
       let unaryStatus;
       let unaryMsg;
       GrpcWebClientBase.setCallback_(
-          stream, (error, response, status, metadata) => {
+          stream, (error, response, status, metadata, trigger) => {
```
sampajano, line 126: could you rename `trigger` to a more explicit name?

Maybe `unaryEndOfStream`?

sampajano (commented):
Hi Good news: the MethodDescriptor getName() change was submitted and didn't raise any alarm in the past week or two. So I think we can safely proceed with this PR 😃

javascript/net/grpc/web/grpcwebclientbase.js
pro-wh: > I'm not sure we should set the `unaryMsg` on an `else {}` case — it seems a bit unsafe to set it before checking nullability (even though it might be ok based on existing callers).

This single callback has to do 5 different jobs in a function.

1. receive an error
2. receive the response value
3. receive the status
4. receive the metadata
5. be informed that we have all pieces needed to give the caller the unary response

So far, it has done this by using a giant `if ...` tree and testing for various arguments' falsiness. Supporting falsy response values would require a way to differentiate job 2 and job 5. This version declares that job 5 has `trigger = true`, while job 2 has `trigger = false`, which overall sticks to the existing theme of "one of the args is truthy, just check which one."

As for how can we ensure it's safe: `setCallback_` is private and the callback handle is kept private, so we have control over what will ever call it, and we can make sure all callers follow this new protocol in this PR.

> Actually.. I'm curious what if you don't make this change? With the existing code, wouldn't the `stream.on('end' ...)` below eventually trigger a `callback(null, null)` call which will fall into this `else {}`clause, which will trigger the `resolve(...)` call?

If we don't change the protocol:

1. A falsy response never gets assigned into `unaryMsg`. I believe our use case is sloppy-tolerant enough not to experience any problems if grpc-web gives `undefined` instead of `0` or `""` or `false` or `null`, but giving the right type is still nice.
2. The `resolve` would be called twice. I think standards compliant Promise implementations are specified to ignore extra calls, so that shouldn't break anything. But that's a relatively more obscure detail, and I think the code is easier for beginners to verify if it only ever calls `resolve` once.

pro-wh: 3. Actually I don't know about the order of calls to this callback. And if I did know the order, I don't know it's an order that gRPC intends to stick to. If job 2 comes before job 4, for example (I vaguely recall seeing something about trailers in gPRC? Maybe grpc-web doesn't support that though.), we'd incorrectly `resolve` before setting the unary metadata.

javascript/net/grpc/web/grpcwebclientbase.js
pro-wh: renamed.

javascript/net/grpc/web/grpcwebclientbase.js
sampajano: Hi! Thanks for the quick reply and sorry for the slow reply.. It took me a while to wrap my head around the code here.. 😅

> As for how can we ensure it's safe: setCallback_ is private and the callback handle is kept private, so we have control over what will ever call it, and we can make sure all callers follow this new protocol in this PR.

What you said makes a lot of sense.. I guess i was kind hoping that there's an explicit signal for setting the response (for better robustness).. but given that the response is falsy here, what you have here is probably the best we can do for now :) (Especially given that the "callback" is not exactly "internal" since we reuse the same function for external callers ([here](https://github.com/grpc/grpc-web/blob/b0ea9c7c45d6f9ea5338fc61a81b47f589a91259/javascript/net/grpc/web/grpcwebclientbase.js#L103-L110)) too..)

> If we don't change the protocol:
> 1. A falsy response never gets assigned into unaryMsg. I believe our use case is sloppy-tolerant enough not to experience any problems if grpc-web gives undefined instead of 0 or "" or false or null, but giving the right type is still nice.
> 2. The resolve would be called twice. I think standards compliant Promise implementations are specified to ignore extra calls, so that shouldn't break anything. But that's a relatively more obscure detail, and I think the code is easier for beginners to verify if it only ever calls resolve once.

Thanks so much for explaining! This makes total sense.. :)

> 3, Actually I don't know about the order of calls to this callback. And if I did know the order, I don't know it's an order that gRPC intends to stick to. If job 2 comes before job 4, for example (I vaguely recall seeing something about trailers in gPRC? Maybe grpc-web doesn't support that though.), we'd incorrectly resolve before setting the unary metadata.

Yeahh i think grpc-web is not using HTTP trailers due to API limitations (see [here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md)) so what you've mentioned shouldn't happen. But thanks for the careful considerations :)

javascript/net/grpc/web/grpcwebclientbase_test.js
```diff
@@ -75,6 +75,33 @@ testSuite({
...
+  async testRpcFalsyResponse() {
+    const xhr = new XhrIo();
+    const client = new GrpcWebClientBase(/* options= */ {}, xhr);
+    const methodDescriptor = createMethodDescriptor((bytes) => {
+      assertElementsEquals(DEFAULT_RPC_RESPONSE_DATA, [].slice.call(bytes));
+      return 0;
+    });
+
+    const response = await new Promise((resolve, reject) => {
+      client.rpcCall(
```
sampajano, line 87: Actually, after scratching my head a bit i realized that the main discussion we had was around "thenableCall" but it's not covered in tests here..

If it's not too difficult, i wonder if you could help add 1-2 thenableCall tests to cover the basic cases (e.g. normal response + false response)? Thanks a lot!!

sampajano (commented):
Thanks so much for the change! Makes total sense! Just a few small things around testing! 😃

javascript/net/grpc/web/grpcwebclientbase.js
pro-wh: here's something we could do, let me know if this is anymore idiomatic:

we still add an extra parameter, but instead of true = "done" and false = "get the response value," we can have true = "get the response value" and false = "done"

javascript/net/grpc/web/grpcwebclientbase_test.js
```diff
@@ -75,6 +75,33 @@ testSuite({
     assertElementsEquals(DEFAULT_UNARY_HEADER_VALUES, Object.values(headers));
   },
 
+  async testRpcFalsyResponse() {
```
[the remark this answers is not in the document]
pro-wh: sure

javascript/net/grpc/web/grpcwebclientbase_test.js
pro-wh: ya let me try

pro-wh: added

javascript/net/grpc/web/grpcwebclientbase.js
pro-wh: pushed a version where instead of a having a unaryEndOfStream flag, there's a unaryResponseReceived flag. i.e., now "there's an explicit signal for setting the response."

added doc comment for the useUnaryResponse  parameter to setCallback_

pro-wh:
I'm getting this when I try to run the tests

> [16:40:38] I/update - chromedriver: file exists /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/chromedriver_93.0.4577.63.zip
[16:40:38] I/update - chromedriver: unzipping chromedriver_93.0.4577.63.zip
[16:40:38] I/update - chromedriver: setting permissions to 0755 for /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/chromedriver_93.0.4577.63
[16:40:38] I/update - chromedriver: chromedriver_93.0.4577.63 up to date
[16:40:38] I/update - selenium standalone: file exists /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/selenium-server-standalone-3.141.59.jar
[16:40:38] I/update - selenium standalone: selenium-server-standalone-3.141.59.jar up to date
[16:40:54] I/launcher - Running 1 instances of WebDriver
[16:40:54] I/direct - Using ChromeDriver directly...
[16:40:54] E/launcher - session not created: This version of ChromeDriver only supports Chrome version 93
Current browser version is 106.0.5249.119 with binary path /usr/bin/google-chrome

is it erroneously using my systemwide chrome rather than its own version for testing?

sampajano:
> I'm getting this when I try to run the tests
> 
> > [16:40:38] I/update - chromedriver: file exists /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/chromedriver_93.0.4577.63.zip
> > [16:40:38] I/update - chromedriver: unzipping chromedriver_93.0.4577.63.zip
> > [16:40:38] I/update - chromedriver: setting permissions to 0755 for /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/chromedriver_93.0.4577.63
> > [16:40:38] I/update - chromedriver: chromedriver_93.0.4577.63 up to date
> > [16:40:38] I/update - selenium standalone: file exists /home/wh0/work/ekiden/grpc-web/packages/grpc-web/node_modules/webdriver-manager/selenium/selenium-server-standalone-3.141.59.jar
> > [16:40:38] I/update - selenium standalone: selenium-server-standalone-3.141.59.jar up to date
> > [16:40:54] I/launcher - Running 1 instances of WebDriver
> > [16:40:54] I/direct - Using ChromeDriver directly...
> > [16:40:54] E/launcher - session not created: This version of ChromeDriver only supports Chrome version 93
> > Current browser version is 106.0.5249.119 with binary path /usr/bin/google-chrome
> 
> is it erroneously using my systemwide chrome rather than its own version for testing?

Aha not sure.. i forgot how i setup my chrome driver.. maybe 93 is a bit too old i can update it soon..

Although, if you run the test using `./scripts/run_basic_tests.sh` then it'll use docker and will probably work.

Could you try that and see if it works? Thanks :)

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -222,9 +222,16 @@ class GrpcWebClientBase {
...
    *     function(?RpcError,?RESPONSE)} callback
-   * @param {boolean} useUnaryResponse
+   * @param {boolean} useUnaryResponse Pass true to have the client make
+   * multiple calls to the callback, using (error, response, status,
+   * metadata, unaryResponseReceived) arguments. One of error, status,
+   * metadata, or unaryResponseReceived will be truthy to indicate which piece
+   * of information the client is providing in that call. After the stream
+   * ends, it will call the callback an additional time with all falsy
+   * arguments. Pass false to have the client make one call to the callback
+   * using (error, response) arguments.
```
sampajano, line 234: nice comments.. appreciate it.. 😃

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -272,11 +279,11 @@ class GrpcWebClientBase {
             message: 'Incomplete response',
           });
         } else {
-          callback(null, responseReceived);
+          callback(null, responseReceived, null, null, /* unaryResponseReceived= */ true);
```
sampajano, line 282: this is where things get a little bit weird.. since in `rpcCall` we're directly taking a user-provided callback.. so this implementation is a bit "leaky".. although, probably (hopefully) not a big deal.. 😄

javascript/net/grpc/web/grpcwebclientbase_test.js
sampajano: wow thanks so much for the new tests. Looks awesome! 😃

javascript/net/grpc/web/grpcwebclientbase.js
```diff
@@ -123,10 +123,10 @@ class GrpcWebClientBase {
       let unaryStatus;
       let unaryMsg;
       GrpcWebClientBase.setCallback_(
-          stream, (error, response, status, metadata) => {
+          stream, (error, response, status, metadata, unaryResponseReceived) => {
             if (error) {
               reject(error);
-            } else if (response) {
+            } else if (unaryResponseReceived) {
```
sampajano, line 129: love this change.. looks much cleaner! Thanks!

sampajano (approved):
Thanks SO much for being patient with the reviews! Change looks fantastic!

Thanks again for working on this change! 😃

sampajano:
I will cut a new release soon (prob. later this week) with this change. Thanks again! 😃

javascript/net/grpc/web/grpcwebclientbase.js
pro-wh: oh good point. I'll open a follow up not to pass the extra flag when useUnaryResponse is off

pro-wh: #1297

sampajano: ah! thanks very much for the follow-up fix! 😃

sampajano:
FYI a new release is cut here :)

https://github.com/grpc/grpc-web/releases/tag/1.4.2

PR #1233, opened 2022-04-30
Remove unused import

### LINKED ISSUE #529: Unused Annotations Imports
My .proto file includes imports that are never used by the generated JS client - eg.
`import google/api/annotations.proto`
`import "protoc-gen-swagger/options/annotations.proto"`

I was able to generate the JS client with `protoc-get-grpc-web`, however, the generated code includes those imports even though they are not used by the client:

```
var google_api_annotations_pb = require('../../google/api/annotations_pb.js')
var protoc$gen$swagger_options_annotations_pb = require('../../protoc-gen-swagger/options/annotations_pb.js')
>>> var `google_protobuf_empty_pb` is declared but its value is never read
```

Is there something in place to prevent unused bindings from being included in the generated code?

Similar issue with grpc-gateway code gen:
https://github.com/grpc-ecosystem/grpc-gateway/issues/310

meling:
This checks which fields are actually used before adding the import alias. That is, some imports are only used by other languages, e.g., option imports such as protopatch for Go. Such imports are not used by Javascript or Typescript, and thus should not be imported. Currently, they result in errors like:

```sh
proto/ag/ag_pb.d.ts:4:30 - error TS2307: Cannot find module '../patch/go_pb' or its corresponding type declarations.

4 import * as patch_go_pb from '../patch/go_pb';
```

Fixes #529.


meling:
Sorry. I'm a bit confused why the already merged commits from master shows up as part of this PR.

sampajano:
> Sorry. I'm a bit confused why the already merged commits from master shows up as part of this PR.

Thanks for the PR again! Maybe try just create a clean branch from master after syncing your personal repo with the upstream? 😃

PR #1235, opened 2022-05-03
Remove unused imports

### LINKED ISSUE #529: Unused Annotations Imports

meling:
This checks which fields are actually used before adding the import alias. That is, some imports are only used by other languages, e.g., option imports such as protopatch for Go. Such imports are not used by Javascript or Typescript, and thus should not be imported. Currently, they result in errors like:

```sh
proto/ag/ag_pb.d.ts:4:30 - error TS2307: Cannot find module '../patch/go_pb' or its corresponding type declarations.

4 import * as patch_go_pb from '../patch/go_pb';
```

Fixes #529.

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -934,15 +934,33 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Print("}\n\n");
 }
 
+// IsUsingImportType returns true if the provided import alias is used by the file.
+bool IsUsingImportType(const FileDescriptor* file, const string alias) {
+  for (int i = 0; i < file->message_type_count(); i++) {
+    const Descriptor* desc = file->message_type(i);
+    for (int j = 0; j < desc->field_count(); j++) {
+      const string field_type = JSFieldType(desc->field(j), file);
+      if (field_type.find(alias) != std::string::npos) {
+        // The import alias is a substring of field_type, thus used by the file
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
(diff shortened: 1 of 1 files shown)


sampajano:
Thanks so much for the fix! 😃

I wonder if you could provide a minimal example in the PR description on when this issue appears and how this fix addresses it?

Also, i wonder if you could consider adding a test for this change? I believe the following place might be a good fit:
https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/test/plugin_test.js

Thanks!! :)

meling:
I'll try to fix these issues. Hopefully later this week.

meling:
Seems unlikely that I'll get to this anytime soon; my brief initial attempts at preparing a test case were not successful (cannot recall the problem, but I think it was related to setup / tooling). While this worked on my particular use case, it is a bit of a hack, so perhaps a more robust solution could be implemented. Closing, but feel free to reopen if anyone else wants to pick this up.

sampajano:
Thanks for the effort and explanation @meling! :)

PR #1246, opened 2022-05-20
make sure envoy is pid 1

ericb-summit:
This ensures signals are handled properly so docker stop responds quickly and doesn't need to wait for time out (10s by default)

By default ENTRYPOINT passed as string (not an array) translates to `/bin/sh -c` followed by the value which means sh is pid 1, not envoy, and `docker stop` hangs for 10s.


sampajano:
Nice trick! Thanks for the PR! Do you mind doing the EasyCLA process? thanks!

ericb-summit:
Done

PR #1252, opened 2022-06-07
Fix Zig setup step in CI

hronro:
Fix the broken CI of building aarch64 binaries.

And I've tested it in my forked repo: https://github.com/hronro/grpc-web/actions/runs/2451424475

### DIFF
--- a/.github/workflows/make-plugin-arm.yml
+++ b/.github/workflows/make-plugin-arm.yml
@@ -17,9 +17,10 @@ jobs:
         submodules: 'recursive'
     - name: Setup Zig
       run: |
-        mkdir /zig
-        curl 'https://ziglang.org/download/0.9.1/zig-linux-x86_64-0.9.1.tar.xz' | tar xJ --strip-components=1 --directory=/zig
-        ln -s /zig/zig /usr/bin/zig
+        mkdir -p $HOME/.local/bin $HOME/.local/zig
+        curl 'https://ziglang.org/download/0.9.1/zig-linux-x86_64-0.9.1.tar.xz' | tar xJ --strip-components=1 --directory=$HOME/.local/zig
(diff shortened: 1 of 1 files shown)


.github/workflows/make-plugin-arm.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         submodules: 'recursive'
     - name: Setup Zig
       run: |
-        mkdir /zig
-        curl 'https://ziglang.org/download/0.9.1/zig-linux-x86_64-0.9.1.tar.xz' | tar xJ --strip-components=1 --directory=/zig
-        ln -s /zig/zig /usr/bin/zig
+        sudo mkdir /zig
+        curl 'https://ziglang.org/download/0.9.1/zig-linux-x86_64-0.9.1.tar.xz' | sudo tar xJ --strip-components=1 --directory=/zig
+        sudo ln -s /zig/zig /usr/bin/zig
```
sampajano, line 22: Thanks for the fix!

Curious does this *have* to happen under root directory? If not, would sudo still be necessary?

hronro: It's about installing for all users (need root permission) or installing for the current user (don't need permission). I think both of them are OK in CI.

sampajano: ahh right.. If you don't mind, i wonder if you could install it for the current user? (in general i'd want to minimize root access unless strictly necessary). Thanks!

hronro: Sure.

sampajano (approved):
thanks so much for the change!! :)

.github/workflows/make-plugin-arm.yml
hronro: lol do you have any simpler solutions?

Issue #1243, opened 2022-05-16
Status Codes are shown in incorrect alphabetical ordering


NoMan2000:
I don't know why, but the StatusCode enums are in the wrong order.

https://github.com/grpc/grpc-web/blob/1779661dded5b918cdd84a216f708b93ebf2384c/packages/grpc-web/index.d.ts#L130

That shows them sorted alphabetically, but they are not actually that way at all.  The actual ordering:

```ts
StatusCode: {
          OK: 0,
          CANCELLED: 1,
          UNKNOWN: 2,
          INVALID_ARGUMENT: 3,
          DEADLINE_EXCEEDED: 4,
          NOT_FOUND: 5,
          ALREADY_EXISTS: 6,
          PERMISSION_DENIED: 7,
          UNAUTHENTICATED: 16,
          RESOURCE_EXHAUSTED: 8,
          FAILED_PRECONDITION: 9,
          ABORTED: 10,
          OUT_OF_RANGE: 11,
          UNIMPLEMENTED: 12,
          INTERNAL: 13,
          UNAVAILABLE: 14,
          DATA_LOSS: 15
        }
 ```

sampajano:
Aha interesting.. thanks for pointing this out!

I'm curious if it causes any build time / runtime issues? Or any other usability issues?

Thanks!

chandraaditya:
Yes it is causing a lot of issues, I did not notice this until now, but I have a check for ```grpcWeb.StatusCode.NOT_FOUND```, it is code ```8``` not ```5``` on ```grpc-web```, therefore when my backend is sending code ```5``` it is being interpreted as ```FAILED_PRECONDITION``` instead of ```NOT_FOUND``` which is the actual code.

sampajano:
@chandraaditya thanks so much for your fix! 

> Yes it is causing a lot of issues, I did not notice this until now, but I have a check for `grpcWeb.StatusCode.NOT_FOUND`, it is code `8` not `5` on `grpc-web`, therefore when my backend is sending code `5` it is being interpreted as `FAILED_PRECONDITION` instead of `NOT_FOUND` which is the actual code.

Also thanks for the comment! Although I'm wonder if this will actually cause a runtime issue or not?

As far as i understand, if you have a check of, for example, `if (statusCode == StatusCode.NOT_FOUND)`, similar to what we have in the TS-client echo demo:

[...]

PR #1279, opened 2022-09-17
Fixes the status codes ordering in typescript definitions

chandraaditya:
In the grpc-web package the status codes are in the wrong order causing them to be interpreted differently, so when comparing status codes that we get from a service it is not the same, this commit fixes that. 

This ordering is causing misinterpretation of error codes:
Wrong ordering

Correct ordering as per gRPC specs: 
Correct order

This fixes issue #1243


sampajano (approved):
Thanks so much for fixing this! I can't believe this was only caught now.. 😅

I think this was a bug introduced in https://github.com/grpc/grpc-web/pull/1139, where i updated `StatusCode` from a namespace to an enum.

I wish we could have a test to catch/verify this, but i don't think we have any e2e TS test yet. Will keep that in mind to add it later.

Thanks so much for the fix again!

PR #1282, opened 2022-09-23
Bump protoc plugin output version to 1.4.0

sampajano:


sampajano:
I believe the interop test failure is a transient failure unrelated to this particular change :)

(they're passing on my local machine too.)

stanley-cheung (commented):
LGTM

PR #1278, opened 2022-09-16
Fix Enum with module in generated TS interface.

sampajano:
Verified with the example in https://github.com/grpc/grpc-web/issues/1271

// enum.proto
```proto
syntax = "proto3";

enum DocEnum {
    DOC_ENUM_PDF = 0;
    DOC_ENUM_HTML = 1;
}
```

// test.proto
```proto
syntax = "proto3";

import "enum.proto";

package Test;

message HelloRequest {
    DocEnum doc = 1;
}
```

Generate:
```
protoc -I=. *.proto --js_out=import_style=commonjs,binary:. --grpc-web_out=import_style=typescript,mode=grpcwebtext:.
```

### Before
```
getDoc(): enum_pbDocEnum;
setDoc(value: enum_pbDocEnum): HelloRequest;
```

### After

```
getDoc(): enum_pb.DocEnum;
setDoc(value: enum_pb.DocEnum): HelloRequest;
```
---

Fixes https://github.com/grpc/grpc-web/issues/1271

Issue #1271, opened 2022-09-04
Missing dot after enum's namespace in typescript file


heggi:
Has 2 proto files:

enum.proto
```proto
syntax = "proto3";

enum DocEnum {
    DOC_ENUM_PDF = 0;
    DOC_ENUM_HTML = 1;
}
```

test.proto
```proto
syntax = "proto3";

import "enum.proto";

package Test;

message HelloRequest {
    DocEnum doc = 1;
}
```

Generated code (test_pb.d.ts)
```ts
import * as jspb from 'google-protobuf'

import * as enum_pb from './enum_pb';


export class HelloRequest extends jspb.Message {
  getDoc(): enum_pbDocEnum;
  setDoc(value: enum_pbDocEnum): HelloRequest;

  serializeBinary(): Uint8Array;
  toObject(includeInstance?: boolean): HelloRequest.AsObject;
  static toObject(includeInstance: boolean, msg: HelloRequest): HelloRequest.AsObject;
  static serializeBinaryToWriter(message: HelloRequest, writer: jspb.BinaryWriter): void;
  static deserializeBinary(bytes: Uint8Array): HelloRequest;
  static deserializeBinaryFromReader(message: HelloRequest, reader: jspb.BinaryReader): HelloRequest;
}

export namespace HelloRequest {
  export type AsObject = {
    doc: enum_pbDocEnum,
  }
}
```

`enum_pbDocEnum` must be `enum_pb.DocEnum`  at 3 places

---
protoc version 3.20.1
protoc-grpc-web 1.3.1
run as `protoc --proto_path=./protos --js_out=import_style=commonjs,binary:gen --grpc-web_out=import_style=typescript,mode=grpcwebtext:gen ./protos/*.proto`

sampajano:
Thanks so much for the bug report!

I'll take a look in the near future! :)

sampajano:
I've taken a look and it seems that the issue is that we forgot to insert a `.` between the package name and the enum name here:

https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/generator/grpc_generator.cc#L338-L340

I'll send a patch soon thanks :)

PR #1286, opened 2022-09-26
Fix duplicate dot in enum name (when "package" is specified)

### LINKED ISSUE #1284: Duplicate dot after enum's namespace in typescript file when using dot in package name
related #1271

#1278 breaks the package name includes dot.

```
$ tree
.
├── child
│   └── enum.proto
└── test.proto

1 directory, 2 files
```

test.proto
```proto
syntax = "proto3";

import "child/enum.proto";

package test;

message HelloRequest {
    child.DocEnum doc = 1;
}
```

child/enum.proto
```proto
syntax = "proto3";

package test.child;

enum DocEnum {
    DOC_ENUM_PDF = 0;
    DOC_ENUM_HTML = 1;
}
```

```
$ protoc-gen-grpc-web --version
protoc-gen-grpc-web 1.4.0
```

```
protoc -I=. *.proto --plugin=/path/to/protoc-gen-js --plugin=/path/to/protoc-gen-grpc-web --js_out=import_style=commonjs:. --grpc-web_out=import_style=commonjs+dts,mode=grpcwebtext:.
```

```ts
import * as jspb from 'google-protobuf'

import * as child_enum_pb from './child/enum_pb';


export class HelloRequest extends jspb.Message {
  getDoc(): child_enum_pb..DocEnum;
  setDoc(value: child_enum_pb..DocEnum): HelloRequest;

  toObject(includeInstance?: boolean): HelloRequest.AsObject;
  static toObject(includeInstance: boolean, msg: HelloRequest): HelloRequest.AsObject;
  static serializeBinaryToWriter(message: HelloRequest, writer: jspb.BinaryWriter): void;
  static deserializeBinary(bytes: Uint8Array): HelloRequest;
  static deserializeBinaryFromReader(message: HelloRequest, reader: jspb.BinaryReader): HelloRequest;
}

export namespace HelloRequest {
  export type AsObject = {
    doc: child_enum_pb..DocEnum,
  }
}
```

sampajano:
## Summary
- Fix a bug introduced in https://github.com/grpc/grpc-web/pull/1278 where an extra dot is inserted when "package" is specified in the enum being imported.

## Changes
- This fix makes `StripPrefixString` remove the leading dot when stripping out a prefix -- and hence will work in both cases (with and without package).
- Also removed a few now redundant logic to remove leading dots.

## Tested
- With both example in #1271 and #1284

Fixes #1284

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -177,12 +177,18 @@ inline bool HasPrefixString(const string& str, const string& prefix) {
          str.compare(0, prefix.size(), prefix) == 0;
 }
 
+// Strips the given prefix from the string, as well as the remaining leading dot
+// if it exists.
 inline string StripPrefixString(const string& str, const string& prefix) {
-  if (HasPrefixString(str, prefix)) {
-    return str.substr(prefix.size());
-  } else {
+  if (!HasPrefixString(str, prefix)) {
     return str;
   }
+
+  string remaining_str = str.substr(prefix.size());
+  if (!remaining_str.empty() && remaining_str[0] == '.') {
+    remaining_str = remaining_str.substr(1);
+  }
+  return remaining_str;
 }
 
 inline bool HasSuffixString(const string& str, const string& suffix) {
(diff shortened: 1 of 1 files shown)


stanley-cheung (approved):
LGTM. Thanks!

Issue #1285, opened 2022-09-26
MethodDescriptor<REQ, RESP>.getName() is not a function


huyungtang:
version: 1.4.0

When calling Request<any, any>.getMethodDescriptor().getName() in unaryInterceptors, all functions of MethodDescriptor<REQ, RESP> disappeared!  In V1.3.1 works fine.

sampajano:
@huyungtang thanks for the report!

Unfortunately that's expected due to the rollback here:
https://github.com/grpc/grpc-web/pull/1199

There's a internal code size increase due to exposing those methods and hence we're leaving them out for now..

If there's a big demand for this we could revisit this decision (by introducing a fork maybe but we generally want to minimize internal & Github divergence).

Thanks again and sorry for inconveniences!

huyungtang:
@sampajano, @stanley-cheung thanks!

But if the definition could be removed at the same time, that would give us lot of help to find out the changes not just in runtime.

sampajano:
> @sampajano, @stanley-cheung thanks!
> 
> But if the definition could be removed at the same time, that would give us lot of help to find out the changes not just in runtime.

That's actually a very good point..

@stanley-cheung Do you think that for now, we can remove `MethodDescriptor` APIs from Typescript interface?

I'm asking since you've made a comment earlier when we first made the fix here: https://github.com/grpc/grpc-web/pull/1160#pullrequestreview-800606603

pro-wh:
> remove `MethodDescriptor` APIs from Typescript interface?

it's used in the `AbstractClientBase` methods though, `thenableCall` et al.

sampajano:
> > remove `MethodDescriptor` APIs from Typescript interface?
> 
> it's used in the `AbstractClientBase` methods though, `thenableCall` et al.

It is indeed used.. But i guess what's being considered here is that whether we can remove the method definitions of `MethodDescriptor` (the class itself can be kept if needed)?

And i guess that generally hinges on whether we would allow custom MethodDescriptors to be provided (like in your use case).. :)

pro-wh:
> remove the method definitions of MethodDescriptor (the class itself can be kept if needed)

I think that's fine with us if you get rid of the types and retain the class. We'll use @ts-ignore as needed, while we export our own typed methods to the users of our library.

> custom MethodDescriptors (like in your use case)

So far there seems to be no conflict here either. We use the same `requestSerializeFn` and `responseDeserializeFn` constructor parameters that the generated grpc code uses.

[...]

PR #1289, opened 2022-09-30
Expose getName() in MethodDescriptor and fix TS definitions.

sampajano:
Backgrounds:

1. MethodDescriptor methods were exposed to match TS definition but later removed in #1199 due to internal code size increase.
2. `getName()` still seemed useful to some users per discussion in https://github.com/grpc/grpc-web/issues/1285

Hence we're re-exposing only the getName() method on MethodDescriptor, hoping it won't cause a code size issue internally and also address some user needs.


stanley-cheung (approved):

dennycd:
lgtm

Issue #270, opened 2018-08-29
Reduce compiled JS file size


stanley-cheung:
As it stands, our [npm package](https://www.npmjs.com/package/grpc-web) `index.js` is 218KiB (which is being compiled by the closure compiler [here](https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/scripts/build.js)).

With the [Echo example echo.proto](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/echo.proto), that has about 7 message types, 1 service, 3 unary methods and 5 streaming methods:
- The [Closure example](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/Makefile#L60) final `compiled.js` compiled by the Closure compiler: 317KiB
 - The [CommonJS example](https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/echo/commonjs-example) final `dist/main.js` compiled by `webpack`: 415KiB
 - The same CommonJS example compiled by `browserify`: 479KiB
 - The CommonJS example [with `.d.ts` typings](https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/echo/ts-example) generated, and then compiled by `webpack`: 499KiB

I tried `babel-minify` and `uglify-js` on some of these JS files and the output are not any particularly smaller.

There should be some other things we can do to further minimize the compiled JS code size.

Reference: #79

color-infinispace:
@stanley-cheung   could you check the size of compiled JS for protobuf messages?

stanley-cheung:
A bit more info, as it stands:
 - the generated protobuf messages `echo_pb.js` produced by `--js_out`: 35KiB
 - the generated gRPC Web stub `echo_grpc_web_pb.js` produced by `--grpc-web_out`: 9KiB

A couple more observations:
 - Just by looking at `compiled.js`, `dist/main.js` produced by the closure compiler, webpack, etc, the compiled JS files are already reasonably minimized. Most variables are already renamed, whitespace removed, etc.
 - From a cursory look, the largest increase of code size is caused by  1. proto messages pulling in `google-protobuf` runtime classes and 2. grpc-web service stub pulling in `goog.net.streams` and closure library code. But again, after those dependencies are pulled in, the resulting code are already reasonably minimized.

wenbozhu:
For the protobuf messages, could you check the compiled.js with out the protobuf  messages.

[...]

PR #1291, opened 2022-10-08
upgrade `google-protobuf` version to fix bundle size issues

sayjeyhi:
Based on this issue: https://github.com/protocolbuffers/protobuf-javascript/issues/124 on google-protobuf the new changes fixed and reduced the bundle size. So here we upgrade `google-protobuf` version to fix bundle size issues.

Screen Shot 1401-07-16 at 12 07 54

### DIFF
--- a/packages/grpc-web/package.json
+++ b/packages/grpc-web/package.json
@@ -30,7 +30,7 @@
     "google-closure-compiler": "~20200224.0.0",
     "google-closure-deps": "~20210601.0.0",
     "google-closure-library": "~20210808.0.0",
-    "google-protobuf": "~3.14.0",
+    "google-protobuf": "~3.21.1",
     "gulp": "~4.0.2",
     "gulp-connect": "~5.7.0",
     "gulp-eval": "~1.0.0",


sayjeyhi:
This will close: https://github.com/grpc/grpc-web/issues/270

sampajano:
@sayjeyhi Hi Jafar, thanks so much for working on the code size issue! Really appreciate your initiative and endeavor here! 😃

Although.. when i just checked bundle size for `google-protobuf` again i've noticed it has reverted to original size (and actually a little bit higher than that)..

Screen Shot 2022-10-11 at 4 52 47 PM

I'm guessing it's due to this PR below:
https://github.com/protocolbuffers/protobuf-javascript/pull/143

I wonder if you have any ideas on the above PR and whether it can be solved?

Thanks a lot!

packages/grpc-web/package.json
```diff
@@ -30,7 +30,7 @@
     "google-closure-compiler": "~20200224.0.0",
     "google-closure-deps": "~20210601.0.0",
     "google-closure-library": "~20210808.0.0",
-    "google-protobuf": "~3.14.0",
+    "google-protobuf": "~3.21.1",
```
sampajano, line 33: btw, since this is only a "devDependencies" and is only used during development, it does not change the final output of the grpc-web npm package (which doesn't depend on google-protobuf directly).

What would make a difference is when developers start depending on `google-protobuf@3.21.1` directly in their client applicaiton, like in the following HelloWorld example:

https://github.com/grpc/grpc-web/blob/b0ea9c7c45d6f9ea5338fc61a81b47f589a91259/net/grpc/gateway/examples/helloworld/package.json#L10

sampajano:
FYI - We're on 3.21.4 already now!

https://github.com/grpc/grpc-web/blob/d48b7a783b6dac62d54b90095b00e1932334b156/packages/grpc-web/package.json#L33

So closing this.

@sayjeyhi Thanks for bringing this up! 😃

PR #1297, opened 2022-10-25
GrpcWebClientBase: clean up !useUnaryResponse callback args

pro-wh:
from #1230, the public `rpcCall` allows users to provide a callback which is documented only to take 2 parameters. avoid passing the extra unaryResponseReceived flag in that case


sampajano (approved):
Thanks so much for cleaning this up! 😃

PR #1310, opened 2023-01-06
Fix TS callback error and response to be nullable

### LINKED ISSUE #1309: Add nullable to RpcError of generated typescript code
I am working with an angular application and eslint. When I enable the `@typescript-eslint/no-unnecessary-condition` rule (see https://typescript-eslint.io/rules/no-unnecessary-condition) it triggers a linting error which is cause by the generated code from the `grpc_generator.cc` file.

Excerpt of the generated code:

```ts
  setFoo(
    request: xyz_api_pb.FooRequest,
    metadata: grpcWeb.Metadata | null,
    callback?: (
      err: grpcWeb.RpcError, // => err is not typed as `grpcWeb.RpcError | null`
      response: xyz_api_pb.FooResponse) => void
  ) {
    // ...
  }
```

Example implementation of the service with angular:

```ts
  public foo(): Promise<void> {
    const request = new FooRequest();

    return new Promise((resolve, reject) => {
      this._apiClient.setFoo(request, {}, (err, _response) => {
        if (err) { // => Unnecessary conditional, value is always truthy 
          reject(err);
        } else {
          resolve();
        }
      });
    });
  }
```

The issue is that `err` is typed as `grpcWeb.RpcError` instead of `grpcWeb.RpcError | null`. The linter checks the if statement and detect it is not necessary, because the generated type states it is always a `grpcWeb.RpcError` , which is not true because it can be `null` if the request succeeds.

Can you adjust the generator to include the nullable type?

sampajano:
IMPORTANT: This will be a **BREAKING change** for some existing users, but hopefully in a good way because now the return type is more accurately reflecting the runtime behavior.

---

In reality, both error and response can be null.

JS Defintion:
https://github.com/grpc/grpc-web/blob/e49389873d887d15ab2870288f620aa2f15b3b85/javascript/net/grpc/web/abstractclientbase.js#L50-L51

Callsites:
https://github.com/grpc/grpc-web/blob/e49389873d887d15ab2870288f620aa2f15b3b85/javascript/net/grpc/web/grpcwebclientbase.js#L249

https://github.com/grpc/grpc-web/blob/e49389873d887d15ab2870288f620aa2f15b3b85/javascript/net/grpc/web/grpcwebclientbase.js#L284

---
Fixes #1309

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -660,8 +660,8 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Print(vars,
                          "request: $input_type$,\n"
                          "metadata: grpcWeb.Metadata | null,\n"
-                         "callback: (err: grpcWeb.RpcError,\n"
-                         "           response: $output_type$) => void): "
+                         "callback: (err: grpcWeb.RpcError | null,\n"
+                         "           response: $output_type$ | null) => void): "
                          "grpcWeb.ClientReadableStream<$output_type$>;\n\n");
           printer->Outdent();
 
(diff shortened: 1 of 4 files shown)


sampajano:
@stanley-cheung Hey Stanley, i think maybe this is a TS bug from day 1. I wonder if there's any concern if we make this fix (probably be breaking for some existing clients but should be easy to fix) and just bump the "minor" version (to 1.5)? Thanks!

sampajano:
Please see https://github.com/grpc/grpc-web/issues/1309#issuecomment-1373045693 — I realized that this fix will create additional unnecessary troubles for callers so it's not idea. Dropping it for now.

Better ideas that does not involve a big breaking change would be welcome!

Issue #1242, opened 2022-05-11
grpc-web not working in a Vite+Typescript app


loicmagne:
I used `protoc -I=. protos/*.proto --js_out=import_style=commonjs,binary:grpcproto_web/ --grpc-web_out=import_style=typescript,mode=grpcweb:grpcproto_web/` to generate my proto file.

Then whenever I try to import proto objects like `import { LoginRequest } from '../protos/login_pb';` I get errors `Uncaught SyntaxError: import not found: LoginRequest`.

I looked at issues [#535](https://github.com/grpc/grpc-web/issues/535) or [vite#3898](https://github.com/vitejs/vite/issues/3898) but nothing works.

Does anyone have a working example using Vite+Typescript with grpc-web ?

Krzysztof-Dziardziel:
I'm having the same issue. Sadly, I couldn't get it to work. I also wasn't able to get Vite+JS to work. Looks like this is some issue with how Vite loads external resources. 😞

Krzysztof-Dziardziel:
@loicmagne You can use @protobuf-ts and `protoc --ts_out ./generated --proto_path "./proto" <protoFilesPaths>` command. It works fine with Vite.

lbensaad:
@Krzysztof-Dziardziel Could you please tell us how you configured protobuf-ts and grpc-web to work with vite?

FredrikMeyer:
@lbensaad I think the suggestion of @Krzysztof-Dziardziel is to replace grpc-web with protobuf-ts. That's what we did at my job (and it works wonderfully - the objects are much easier to work with, they are just regular JS objects)

wirekang:
> @loicmagne You can use `@protobuf-ts` and `protoc --ts_out ./generated --proto_path "./proto" <protoFilesPaths>` command. It works fine with Vite.

This is best answer. We replace grpc-web with protobuf-ts, and every problem have been solved.

lbensaad:
Thanks @FredrikMeyer, I read it above, but I thought that `protobuf-ts` is only for messages and must use `grpc-web` for RPC. Now I also use `protobuf-ts` for RPC and works perfectly. The key word is **_replace_**. Thanks again for mentioning it.

barrelful:
@sampajano is there any plan to fix grpc-web? I just used the command from the README.md for TypeScript:

```
protoc -I=$DIR echo.proto \
  --js_out=import_style=commonjs,binary:$OUT_DIR \
[...]

PR #1314, opened 2023-01-19
Fix generated Typescript file documentation.

sampajano:
As discovered in https://github.com/grpc/grpc-web/issues/1242#issuecomment-1383860124


sampajano:
@barrelful FYI the documentation was wrong. There is a generated TS file but with a different name than what's documented.. :)

stanley-cheung (approved):
Thanks for the fix!

PR #1313, opened 2023-01-09
Update ES6 imports with comment about corresponding proto import path.

reddaly:
Replaces https://github.com/grpc/grpc-web/pull/1311, which I created with my non-corporate account. I can use the corporate account to meet the CLA requirement.

I was experiencing a long import path in the generated code that doesn't work. Outputting the proto file that corresponds to a JS import is helpful for both diagnosing the issue and writing a wrapper for altering the import path to my satisfaction.

Before:

```js
import * as github_com_gonzojive_rules_ts_proto_example_prefix_greeting_pb from '../../../../../github.com/gonzojive/rules_ts_proto/example/prefix/greeting_pb';
```

After:

```js
import * as github_com_gonzojive_rules_ts_proto_example_prefix_greeting_pb from '../../../../../github.com/gonzojive/rules_ts_proto/example/prefix/greeting_pb'; // proto import: "github.com/gonzojive/rules_ts_proto/example/prefix/greeting.proto"
```

PR #1330, opened 2023-04-15
Update ES6 .d.ts imports with comment about corresponding proto import path

gonzojive:
https://github.com/grpc/grpc-web/pull/1313 updated the code generator to print a ` // proto import: "foo/bar.proto"` suffix for each import statement for the generated gRPC-web client code. This commit updates ensures the .d.ts file output by grpc-web contains the same type of comments.


sampajano (approved):
Thanks so much for the change!

Sorry for the delay too — not sure why this slipped through my list!

gonzojive:
Thanks. I just merged in the latest changes. I think @sampajano will need to trigger the tests and complete the merge.

sampajano:
> Thanks. I just merged in the latest changes. I think @sampajano will need to trigger the tests and complete the merge.

@gonzojive Yes will do! Thanks again for the change!

Issue #1349, opened 2023-07-27
TypeScript example fails to start


Kaszanas:
Looking at the main `README` file I was pointed towards: https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/echo/ts-example

When running the specified command:

```
docker-compose up --build node-server envoy ts-client
```

The following happens:

```
[+] Running 5/5
 ✔ Network grpc-web-master_default          Created                                                          0.1s 
 ✔ Container grpc-web-master-prereqs-1      Created                                                          1.8s 
 ✔ Container grpc-web-master-node-server-1  Created                                                          0.1s 
 ✔ Container grpc-web-master-ts-client-1    Created                                                          0.2s 
 ✔ Container grpc-web-master-envoy-1        Created                                                          0.1s 
Attaching to grpc-web-master-envoy-1, grpc-web-master-node-server-1, grpc-web-master-ts-client-1
grpc-web-master-ts-client-1    | node:internal/modules/cjs/loader:1085
grpc-web-master-ts-client-1    |   throw err;
grpc-web-master-ts-client-1    |   ^
grpc-web-master-ts-client-1    | 
grpc-web-master-ts-client-1    | Error: Cannot find module '/var/www/html/python'
grpc-web-master-ts-client-1    |     at Module._resolveFilename (node:internal/modules/cjs/loader:1082:15)        
grpc-web-master-ts-client-1    |     at Module._load (node:internal/modules/cjs/loader:928:27)
grpc-web-master-ts-client-1    |     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
grpc-web-master-ts-client-1    |     at node:internal/main/run_main_module:23:47 {
grpc-web-master-ts-client-1
```

This is without any changes to the code and on `master` branch, which is the only one available.

sampajano:
Hi thanks for the report! I'm able to reproduce this.. Will work on a fix soon thanks!

sampajano:
@Kaszanas Thanks so much for reporting the issue. It should be fixed now! Please try again thanks! 😃

Kaszanas:
Will try to run the fix as soon as possible, and let you know if it works 😉.

Kaszanas:
Unfortunately, I think it did not help:

```
 => ERROR [prereqs prepare 9/9] RUN ./scripts/init_submodules.sh                                            0.5s 
------
[...]

PR #1352, opened 2023-07-31
Use cloned init_submodules.sh in prereqs (#1349)

Kaszanas:
The problem with building the docker-compose example was due to the additional line copying the `init_submodule.sh` script from the context. The line just above was cloning the repository though, so there is no need to COPY that script.

This is supposed to be solving the issue: https://github.com/grpc/grpc-web/issues/1349

While solving this, another issue was discovered:

### DIFF
--- a/net/grpc/gateway/docker/prereqs/Dockerfile
+++ b/net/grpc/gateway/docker/prereqs/Dockerfile
@@ -42,7 +42,6 @@ RUN wget -nv -O buildifier \
 WORKDIR /github/grpc-web
 
 RUN git clone https://github.com/grpc/grpc-web .
-COPY ./scripts/init_submodules.sh ./scripts/
 RUN ./scripts/init_submodules.sh
 
 


sampajano (commented):
@Kaszanas Thanks for the PR!

I believe this `COPY` line you're removing has its purpose — it's for using a local `init_submodules.sh` instead of the one from github.

So i'd like to understand the problem better before proceeding with this PR.

---

So for context, i believe you're seeing the issue mentioned [here](https://github.com/grpc/grpc-web/issues/1349#issuecomment-1659254109) before this PR.

Do you mind sharing your docker version where you're seeing this error?

I'm on Docker version 20.10.8 (on Mac) and i'm not seeing the same error.

Thanks!

Kaszanas:
Sure thing,

Yes there's one closed issue, and another one which explores further another error down the line.

I'm running Docker on Windows. As I have been performing a clean install recently I have the most recent Docker version (for a more precise answer I'd have to verify).

There may be a couple of things to check:
- is the context passed correctly from the docker-compose vs dockerfile?

sampajano:
> Sure thing,
> 
> Yes there's one closed issue, and another one which explores further another error down the line.

Yes.. have understood about that :)

> I'm running Docker on Windows. As I have been performing a clean install recently I have the most recent Docker version (for a more precise answer I'd have to verify).

Thanks for clarifying on the platform!

I'm running Docker `20.10.8` on Mac, and i was not able to see the same error you did.

That is, i can successfully build using this command:
```
docker-compose build prereqs
```

I guess that's the same command you're running and it's failing?

I'm wondering what is the difference here.

> There may be a couple of things to check:
> 
> * is the context passed correctly from the docker-compose vs dockerfile?

I'm not sure what exactly to check here. Could you clarify? Do you have a theory on what might be missing?

Kaszanas:
I'm running the docker compose command which is supposed to build and run all images I guess. This is following the instructions from the TypeScript example. I will see if the docker build command by itself runs as intended.

sampajano:
> I'm running the docker compose command which is supposed to build and run all images I guess. This is following the instructions from the TypeScript example. I will see if the docker build command by itself runs as intended.

I was running docker compose as well.. But it was succeeding for me..

The command i used is:

```
docker-compose build prereqs
```

sampajano:
Hi :) I'm closing this due to inactivity. Feel free to open another PR when needed. Thanks! :)

Kaszanas:
No problem @sampajano at some point this may resurface for someone else ;)

sampajano:
@Kaszanas thank you :) yes definitely can. We'll address it when it comes :)

PR #1354, opened 2023-08-17
Update echo.proto, remark unsupported method

icoco:
Client side streaming and Bidi streaming are not supported at the moment.


sampajano (commented):
@icoco Hi Thanks so much for this change!

This makes total sense!

But actually, would you mind completely removing them (both method and request/response definitions), since it'll make things every more obvious? 😃

Thanks!!

sampajano (approved):
@icoco Haha appreciate for the optimism! :D Sounds good to me.. 😄

Issue #1356, opened 2023-08-24
Getting {code: 2, message: 'Incomplete response'} code :  2 message :  "Incomplete response" [[Prototype]] :  Object


SimulShift:
I am sending the Status.OK code but with a simple ping pong example, but the browser always gets a code: 2 "Incomplete Message". Everything else is working. If I set the error as null, the RPC gets called correctly and everything functions. grpc-web just seems to have an issue decoding the error messages. Maybe my version of grpc-js is too high ( I am using grpc-js 1.9.1 but the example is using 1.1.8). I tried downgrading but got errors in the node_modules

```
export const pingPongServiceHandlers: PingPongServiceHandlers = {
  ping: (call, callback) => {
    console.log('ping')
    callback(
      {
        code: Status.OK,
        message: 'pong',
        details: 'pong',
        metadata: call.metadata,
        name: 'pong',
      },
      null,
    )
  },
}
```

SimulShift:
after more digging it only happens when using Status.OK and also Status.UNAUTHENTICATED. seems grpc-web does not have code 16 yet (goes to 15). and Status.OK is code: 0 that may be causing some weird typescript conversion issues whewre its interpreting 0 as falsey rather than a status code

sampajano:
@SimulShift Thanks for the report!

It is surprising for me to hear the basic case doesn't work.. since we should have test coverages for them..

I wonder if you are able to reproduce following our Typescript demo here (either using Docker or manual run)?
https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/echo/ts-example#readme

If you are able to, would you mind creating a test PR demo'ing how to reproduce?

Thanks!

SimulShift:
hi @sampajano Here is a [minimum reproducible example](https://github.com/SimulShift/hello_word_status_bug) I just copied the hello world example from this repo and after doing the setup I just set the error code to 0 in the hello reply 

```
function doSayHello(call, callback) {
  callback({ code: 0, message: "Not found" });
}

```
that gives the following response on the browser:

![image](https://github.com/grpc/grpc-web/assets/62919684/9b9842d5-eb3e-4210-915a-93a2a9767102)

SimulShift:
@sampajano sure thing, here is a PR

https://github.com/grpc/grpc-web/pull/1358

sampajano:
@SimulShift Thanks so much for providing the PR example! (And sorry for the delay in response!)

I was able to reproduce this issue.

[...]

PR #1358, opened 2023-09-02
Status 0 causes error code: 2 unknown error on the client

SimulShift:
Status 0 causes error code: 2 unknown error on the client

image

I just changed the hello reply to return a code 0 instead to reproduce the error.


sampajano:
Closing this PR. Please see https://github.com/grpc/grpc-web/issues/1356 for context. thanks :)

Issue #1368, opened 2023-10-05
TypeScript client method signatures should mark metadata as optional


andrewmbenton:
This is an inverse issue to https://github.com/grpc/grpc-web/issues/408

There are a couple of method signatures in the generated TypeScript client that should mark the `metadata` parameter as optional, to reduce explicitly passing unnecessary `null` values:

https://github.com/grpc/grpc-web/blob/0ec55aaad784e114bcb391b0b33ac4fd2e83e3f2/javascript/net/grpc/web/generator/grpc_generator.cc#L652-L655

https://github.com/grpc/grpc-web/blob/0ec55aaad784e114bcb391b0b33ac4fd2e83e3f2/javascript/net/grpc/web/generator/grpc_generator.cc#L670-L674

Making the `metadata` parameter optional in the method returning a `Promise` would bring the TypeScript client closer in line with the `PromiseClient` interface available when generating with `import_style=commonjs+dts`.

PR #1369, opened 2023-10-05
feat(typescript): mark some `metadata` parameters as optional

andrewmbenton:
Resolves https://github.com/grpc/grpc-web/issues/1368

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -653,7 +653,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Indent();
           printer->Print(vars,
                          "request: $input_type$,\n"
-                         "metadata: grpcWeb.Metadata | null): "
+                         "metadata?: grpcWeb.Metadata | null): "
                          "$promise$<$output_type$>;\n\n");
           printer->Outdent();
 
@@ -671,7 +671,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Indent();
           printer->Print(vars,
                          "request: $input_type$,\n"
-                         "metadata: grpcWeb.Metadata | null,\n"
+                         "metadata?: grpcWeb.Metadata | null,\n"
                          "callback?: (err: grpcWeb.RpcError,\n"
                          "           response: $output_type$) => void) {\n");
           printer->Print(vars, "if (callback !== undefined) {\n");


javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -671,7 +671,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Indent();
           printer->Print(vars,
                          "request: $input_type$,\n"
-                         "metadata: grpcWeb.Metadata | null,\n"
+                         "metadata?: grpcWeb.Metadata | null,\n"
```
sampajano, line 674: You've mentioned this in the associated issue (which i copied to this PR too :))

> Making these parameters optional I believe would bring the TypeScript client closer in line with the `PromiseClient` interface available when generating with `import_style=commonjs+dts`.

I believe this is true for the other change! But not this one.

It seems that `callback?` has only 1 occurrence the whole file, and is a unique feature in TS codegen :)

I don't mind this change also, but maybe you could consider rewording your comment (and PR description now :)) to note that it only applies to the other change? Thanks!

sampajano (approved):
@andrewmbenton Thanks so much for this change (and sorry for the delay)! This makes a lot of sense! 😃

Just one comment! Will merge once it's addressed/acked, and when tests passes!

Thanks again for the contribution! :)

javascript/net/grpc/web/generator/grpc_generator.cc
andrewmbenton: I will reword my comments. You are right that the `PromiseClient` generated with `import_style=commonjs+dts` doesn't seem to have any generated methods with a `callback` parameter.

For true parity with the `PromiseClient` I would remove the `null` from the `metadata` param's type, but that would break user code that passes `null` so I didn't propose that change.

I am also happy to revert the change to the function that doesn't return a `Promise` if that's preferable.

sampajano: > For true parity with the PromiseClient I would remove the null from the metadata param's type, but that would break user code that passes null so I didn't propose that change.

Definitely! Appreciate the caution and i'd agree!

> I am also happy to revert the change to the function that doesn't return a Promise if that's preferable.

Yeah i think that's an option but i don't see this making things more inconsistent than they already were (at least my glimpse of it) so i think this is still a good thing to keep 😃

Thanks for offering tho!

sampajano (approved):
Thanks so much for the change again! Much appreciated!

PR #1397, opened 2024-01-23
Update http2_protocol_options pattern

dimo414:
In my Envoy logs I see a number of entries like this:

> `[2024-01-22 15:44:47.698][620][warning][misc] [source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.config.cluster.v3.Cluster Using deprecated option 'envoy.config.cluster.v3.Cluster.http2_protocol_options' from file cluster.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.`

I believe this stems from https://github.com/envoyproxy/envoy/pull/14079 (@alyssawilk) but the examples and tests in this repo haven't been updated. I lifted [this diff](https://github.com/envoyproxy/envoy/pull/14079/files#diff-fa199c63de1a5dba807902ac586a6108a08913f78c6ce0040211f8015c66201e) from that PR and applied to all mentions of this string in this repo. I verified this change works in my Envoy application.

### DIFF
--- a/net/grpc/gateway/examples/echo/envoy.yaml
+++ b/net/grpc/gateway/examples/echo/envoy.yaml
@@ -48,7 +48,12 @@ static_resources:
     - name: echo_service
       connect_timeout: 0.25s
       type: logical_dns
-      http2_protocol_options: {}
+      # HTTP/2 support
+      typed_extension_protocol_options:
+        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+          "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+          explicit_http_config:
+            http2_protocol_options: {}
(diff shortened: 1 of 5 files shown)


alyssawilk:
yeah so IMO it's better to use new config but I've also had conversations with the envoy API shepherds about removing those warnings as we're no longer removing deprecated config.  they suggested keeping the warnings last time I checked but maybe they're open to it now?  either way thanks for the fix!

sampajano (approved):
Thank you very much for help updating this! 😃

Issue #1391, opened 2023-12-08
mode=grpcweb support server streaming in http2 actually


jmzwcn:
mode=grpcwebtext: The default generated code sends the payload in the grpc-web-text format.

Content-type: application/grpc-web-text
Payload are base64-encoded.
Both unary and server streaming calls are supported.
mode=grpcweb: A binary protobuf format is also supported.

Content-type: application/grpc-web+proto
Payload are in the binary protobuf format.
**Only unary calls are supported for now. ==> Both unary and server streaming calls are supported.**

sampajano:
Thanks so much for the info!!

See https://github.com/grpc/grpc-web/discussions/1322#discussioncomment-7804499

I'd like to have one more confirmation here before i update our docs! 😃

jmzwcn:
furthermore, I find: based on grpc-web, we could directly send pb msg to grpc-server without any proxy.

https://github.com/emart-io/cross/blob/main/zwan/internal/cmd/main.go#L49

sampajano:
> Furthermore, I find based on grpc-web, we could directly send pb msg to grpc-server without any proxy.
> 
> https://github.com/emart-io/cross/blob/main/zwan/internal/cmd/main.go#L49

@jmzwcn Aha interesting! Good to note!! I guess grpc-server would have recently added grpc-web support..?

jmzwcn:
No, if server stream is always sending without stop,  grpc-web client will don't fetch those stream.

Issue #1396, opened 2024-01-11
Status of streaming with the binary wire format


dimo414:
The readme section for `mode=grpcweb` notes that "Only unary calls are supported for now". This line [dates back to 2018](https://github.com/grpc/grpc-web/pull/251). Is this still correct? Are there any plans to resolve this wire format incompatibility?

sampajano:
Yeah i believe this is still true. And unfortunately no, nothing is planned around improving streaming support in binary mode.. :)

sampajano:
I'll close this issue for now. Please feel free to open another feature request issue if there's a good reason why it would be beneficial to add this support. thanks! :)

dimo414:
Is the expectation then that we should just always use the text format? It's a really surprising limitation of the binary format that it lacks features the text format supports.

At a minimum it would be helpful to remove the "for now" if there are not actually any plans to resolve this discrepancy.

sampajano:
> Is the expectation then that we should just always use the text format? It's a really surprising limitation of the binary format that it lacks features the text format supports.

For streaming i guess so. I was not involved in the earlier design/implementations, but afaiu there's some limitations with how messages are encoded/decoded which makes streaming decode not feasible when binary format is used..

> At a minimum it would be helpful to remove the "for now" if there are not actually any plans to resolve this discrepancy.

That's quite fair! Let me remove that "for now" text then! :)

PR #1403, opened 2024-02-09
Update document to mention Nginx server-streaming support

sampajano:
Discussed in following threads:

https://github.com/grpc/grpc-web/issues/1391
https://github.com/grpc/grpc-web/discussions/1322

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ frameworks for languages such as Python, Java, and Node. For details, see the
 ## Streaming Support
 gRPC-web currently supports 2 RPC modes:
 - Unary RPCs ([example](#make-a-unary-rpc-call))
-- Server-side Streaming RPCs ([example](#server-side-streaming)) (NOTE: Only when [`grpcwebtext`](#wire-format-mode) mode is used.)
+- Server-side Streaming RPCs ([example](#server-side-streaming)) (NOTE: Only when [`grpcwebtext`](#wire-format-mode) mode is used, or when [Nginx](#ecosystem) is used)
 
 Client-side and Bi-directional streaming is not currently supported (see [streaming roadmap](doc/streaming-roadmap.md)).
 
@@ -123,7 +123,7 @@ For more information about the gRPC-web wire format, see the
 
   - `Content-type: application/grpc-web+proto`
   - Payload are in the binary protobuf format.
-  - Only unary calls are supported.
+  - Only unary calls are supported, unless when [Nginx](#ecosystem) is used.
 
 ## How It Works
 
@@ -337,15 +337,16 @@ Multiple proxies support the gRPC-web protocol.
 	$ docker-compose up -d node-server envoy commonjs-client
 	```
 
-2. You can also try the [gRPC-web Go proxy][].
+2. [Nginx](https://www.nginx.com/) has a grpc-web module ([doc](https://nginx.org/en/docs/http/ngx_http_grpc_module.html), [announcement](https://www.nginx.com/blog/nginx-1-13-10-grpc/)). It seems to work with simple configs, according to [user feedbacks](https://github.com/grpc/grpc-web/discussions/1322)), even with server-streaming support ([user feedback](https://github.com/grpc/grpc-web/issues/1391)).
+
+3. You can also try the [gRPC-web Go proxy][].
 
 	```sh
 	$ docker-compose up -d node-server grpcwebproxy binary-client
 	```
 
-3. Apache [APISIX](https://apisix.apache.org/) has also added grpc-web support, and more details can be found [here](https://apisix.apache.org/blog/2022/01/25/apisix-grpc-web-integration/).
+4. Apache [APISIX](https://apisix.apache.org/) has also added grpc-web support, and more details can be found [here](https://apisix.apache.org/blog/2022/01/25/apisix-grpc-web-integration/).
 
-4. [Nginx](https://www.nginx.com/) has a grpc-web module ([doc](https://nginx.org/en/docs/http/ngx_http_grpc_module.html), [announcement](https://www.nginx.com/blog/nginx-1-13-10-grpc/))), and seems to work with simple configs, according to user [feedback](https://github.com/grpc/grpc-web/discussions/1322).
 
 ### Server Frameworks with gRPC-Web support
 - [Armeria (JVM)](https://armeria.dev/docs/server-grpc/#grpc-web)


README.md
```diff
@@ -13,7 +13,7 @@ frameworks for languages such as Python, Java, and Node. For details, see the
 ## Streaming Support
 gRPC-web currently supports 2 RPC modes:
 - Unary RPCs ([example](#make-a-unary-rpc-call))
-- Server-side Streaming RPCs ([example](#server-side-streaming)) (NOTE: Only when [`grpcwebtext`](#wire-format-mode) mode is used.)
+- Server-side Streaming RPCs ([example](#server-side-streaming)) (NOTE: Only when [`grpcwebtext`](#wire-format-mode) mode is used, or when [Nginx](#ecosystem) is used)
```
diamond-504, line 16: This is a surprising / confusing note. Why does Envoy, the gRPC-web recommended proxy, not support server-side streaming in binary, if Nginx is able to?

README.md
```diff
@@ -337,15 +337,16 @@ Multiple proxies support the gRPC-web protocol.
 	$ docker-compose up -d node-server envoy commonjs-client
 	```
 
-2. You can also try the [gRPC-web Go proxy][].
+2. [Nginx](https://www.nginx.com/) has a grpc-web module ([doc](https://nginx.org/en/docs/http/ngx_http_grpc_module.html), [announcement](https://www.nginx.com/blog/nginx-1-13-10-grpc/)). It seems to work with simple configs, according to [user feedbacks](https://github.com/grpc/grpc-web/discussions/1322)), even with server-streaming support ([user feedback](https://github.com/grpc/grpc-web/issues/1391)).
```
diamond-504, line 340: Neither of these pages mention gRPC-web - are we sure they are referring to gRPC-web and not gRPC?

README.md
dimo414, line 16: This is a surprising / confusing note. Why does Envoy, the gRPC-web recommended proxy, not support server-side streaming in binary, if Nginx is able to?

README.md
dimo414, line 340: Neither of these nginx pages mention gRPC-web - are we sure they are referring to gRPC-web and not gRPC?

README.md
sampajano: Yeah, i can agree.. I was just stating the information i gathered, but not trying to reason if it's "expected" or not.. :)

README.md
sampajano: Hmm, you're right.. i think that's a very good point.. Somehow i thought they were — I probably misread initially. Maybe they are not.. 

In any case, there are 2 users reporting grpc-web is working, including the OP and this [confirmation](https://github.com/grpc/grpc-web/discussions/1322#discussioncomment-8318790), and hence i changed the doc.

Maybe i should wait for one extra confirmation on the binary streaming case, but the OP sounded pretty sure in https://github.com/grpc/grpc-web/issues/1391 so i trusted his judgements :)

dimo414: I'm pretty suspicious these threads are not consistently distinguishing gRPC traffic from gRPC-web. For example [`application/grpc`](https://github.com/grpc/grpc-web/discussions/1322#discussioncomment-8318790) is [clearly not gRPC-web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2). Furthermore, it seems unlikely that Nginx has _somehow_ managed to enable streaming support for the gRPC-web binary format when [your understanding was the binary format is incapable of supporting server-side streaming](https://github.com/grpc/grpc-web/issues/1396#issuecomment-1920218318).

More to the point, it's _impossible_ for Nginx to have enabled streaming support for the gRPC-web binary protocol without collaborating with the gRPC-web project, because the protocol spec directly says this is not supported. So if they have implemented some sort of streaming support it's not to spec.

But my bet (especially given the lack of any mention of gRPC-web in their documentation) is Nginx supports _gRPC_, and people are conflating it with gRPC-web because it's a "web" server that's able to proxy gRPC traffic.

README.md
dimo414: The phrasing here is still confusing, even if it's true - it says only the text format supports streaming, but also that Nginx actually supports streaming regardless of format. _If true_ it would be clearer to say that streaming conceptually works with both formats but the Envoy implementation specifically only supports streaming with the text format.

sampajano: Yeah i agree with you.. i didn't call out Envoy specifically because it's [not the only](https://github.com/grpc/grpc-web?tab=readme-ov-file#proxy-interoperability) proxy that supports grpc web.

Let's get a conclusion on the other comment you have, and then i'll update this text.

README.md
sampajano: > I'm pretty suspicious these threads are not consistently distinguishing gRPC traffic from gRPC-web. For example [`application/grpc`](https://github.com/grpc/grpc-web/discussions/1322#discussioncomment-8318790) is [clearly not gRPC-web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2).

Yeah i can see your suspicion. I don't deny them.

> Furthermore, it seems unlikely that Nginx has _somehow_ managed to enable streaming support for the gRPC-web binary format when [your understanding was the binary format is incapable of supporting server-side streaming](https://github.com/grpc/grpc-web/issues/1396#issuecomment-1920218318).

This is a very good point.. thanks for pointing it out :)

> More to the point, it's _impossible_ for Nginx to have enabled streaming support for the gRPC-web binary format without collaborating with the gRPC-web project, because the protocol spec directly says this is not supported. So if they have implemented some sort of streaming support it's not to spec.

Yes another very good point.

> But my bet (especially given the lack of any mention of gRPC-web in their documentation) is Nginx supports _gRPC_, and people are conflating it with gRPC-web because it's a "web" server that's able to proxy gRPC traffic.

Yes it's definitely a possibility. Let me loop in the original commenters.

---

@jmzwcn @zhumazhenis Hi! Thanks for previously participating in the [discussion](https://github.com/grpc/grpc-web/discussions/1322) on nginx support.

There's a question now on the validity of the conclusions as above.

Could you help responding to it if there are some misunderstandings?

Thanks in advance!

dimo414: @sampajano any objection to reverting these notes? At best they're unclear and IMO it would be better to remove them and hash out exactly what is known before re-adding them.

sampajano: OK! Since the original commenters are not active at this point, i guess we can revert until upon further discussions :)

Do you want to create a PR to revert this one? Or if you prefer i can do it too.

Thanks! :)

dimo414: Thanks, just cut #1408

PR #1404, opened 2024-02-12
Update interop-test-descriptions.md

tsegismont:
The proto definition link leads to a 404

The actual location of the file seems to be https://github.com/grpc/grpc/blob/master/src/proto/grpc/testing/test.proto

### DIFF
--- a/doc/interop-test-descriptions.md
+++ b/doc/interop-test-descriptions.md
@@ -4,7 +4,7 @@ gRPC-Web Interop Tests
 This document describes the set of tests any gRPC-Web clients or proxies need
 to implement. The proto definition for the messages and RPCs we are using for
 the tests can be found
-[here](src/proto/grpc/testing/test.proto).
+[here](https://github.com/grpc/grpc/blob/master/src/proto/grpc/testing/test.proto).
 
 The canonical set of interop tests was defined in the main
 [grpc/grpc repo](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md).


sampajano:
Wonderful! Thanks for the fix! Do you mind complete the EasyCLA process btw? Thanks! 😊

sampajano (approved):
Thanks for the fix! 😊

tsegismont:
Thank you @sampajano for the quick reply. I've authorized the EasyCLA app and the check passed.

I was reading this document because I'm working on adding gRPC Web support to the vert-grpc (I'm an Eclipse Vert.x committer).

It seems I should be able to use the test suite described in https://github.com/grpc/grpc-web/tree/master/test/interop, is that right?

sampajano:
> Thank you @sampajano for the quick reply. I've authorized the EasyCLA app and the check passed.
> 
> I was reading this document because I'm working on adding gRPC Web support to the vert-grpc (I'm an Eclipse Vert.x committer).
> 
> It seems I should be able to use the test suite described in https://github.com/grpc/grpc-web/tree/master/test/interop, is that right?

@tsegismont You're very welcome!!

Oh thanks for adding this support!

Yeah on a quick glance i assume that could work!

From our dockerfile here:
https://github.com/grpc/grpc-web/blob/55b9218d2bf435becbb234b4251e9ba860b17d6c/net/grpc/gateway/docker/node_interop_server/Dockerfile#L37

It seems that we're using this interop-server by grpc-node:
https://github.com/grpc/grpc-node/blob/master/test/interop/interop_server.js

So maybe you'd have to implement a similar server and the same test can likely be run on there too 😃

tsegismont:
Hi @sampajano , the implementation of gRPC-Web in Vert.x is [almost ready](https://github.com/eclipse-vertx/vertx-grpc/pull/89).

I've implemented a backend for the interop tests and used the [grpcweb/prereqs](https://hub.docker.com/r/grpcweb/prereqs) image to run them in a container. I noticed on Docker Hub that the image was updated two years ago. Is that expected? Or should I instead create the image using a fork of this repo during CI builds?

sampajano:
> Hi @sampajano , the implementation of gRPC-Web in Vert.x is https://github.com/eclipse-vertx/vertx-grpc/pull/89.

Oh wow! Amazing! Thanks for the good news! 😃

>  I noticed on Docker Hub that the image was updated two years ago. Is that expected?

Hi! Thanks for pointing it out! I mean kinda yes and no.. Yes because I've not been pushing them.. No because it's not intentional.. 😄 I will probably try pushing them soon (like during next release).

For now, could you use `docker-compose build` to build them locally during testing? (which is what the testing script currently [does](https://github.com/grpc/grpc-web/blob/a639b4cf2611de2b68883571787083b73cf61f5e/scripts/run_interop_tests.sh#L45)). Would that work for you?

Thanks!

tsegismont:
> Oh wow! Amazing! Thanks for the good news! 😃

Thank you!



> For now, could you use `docker-compose build` to build them locally during testing? (which is what the testing script currently [does](https://github.com/grpc/grpc-web/blob/a639b4cf2611de2b68883571787083b73cf61f5e/scripts/run_interop_tests.sh#L45)). Would that work for you?

I changed the test so that a docker image is built from a fresh checkout of this repo: https://github.com/eclipse-vertx/vertx-grpc/pull/89/commits/35b180f7659033c42c96cb3009a42e886c009196

I chose to build only the `prereqs` container because in the Vert.x repo, we don't need the interop server.

tsegismont:
No worries, thank you for taking some time to answer my questions!

sampajano:
You're very welcome! 😃

PR #1408, opened 2024-03-06
Revert "Update document to mention Nginx server-streaming support"

dimo414:
Reverts grpc/grpc-web#1403 - it's not clear (and, I suspect, unlikely) that Nginx actually supports gRPC-web.


sampajano (approved):
@dimo414 Thanks for help pushing for more accurate documentation! 😃

PR #1418, opened 2024-04-02
Use `docker compose` instead of `docker-compose`

sampajano:
The latter seems deprecated :)


sampajano:
Seems our CI runtime isn't prime for `docker compose` yet :)

Issue #1331, opened 2023-04-18
Typescript: Property 'unaryCall' does not exist on type 'AbstractClientBase'


GabrieleGervasiPEL:
Running protoc with grp-web plugin (1.4.2) generated a file *ServiceClientPb.ts. Such file contains a call to this.client_.unaryCall.
Meantime, npm module grp-web (1.4.2) doesn't contain such method in AbstractClientBase class.

Protoc configuration for plugin:
--grpc-web_out=import_style=typescript,mode=grpcweb

sampajano:
@GabrieleGervasiPEL Hi there :) Thanks for bringing up this question!

You're indeed right that `unaryCall` is not exposed on the `AbstractClientBase` class..
https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/index.d.ts

Although, they do exist on the actual implementation of the class, which should work when it's actually called..
https://github.com/grpc/grpc-web/blob/8ab32b945ca0530222593127e9431a455eff24a8/javascript/net/grpc/web/grpcwebclientbase.js#L158-L161

(As of now,  `unaryCall()` is just an alias for `thenableCall()`, where it simply [calls](https://github.com/grpc/grpc-web/blob/8ab32b945ca0530222593127e9431a455eff24a8/javascript/net/grpc/web/grpcwebclientbase.js#L158-L161) the latter.)

I agree this is not ideal, and I will explore ways to close the gap (e.g. https://github.com/grpc/grpc-web/pull/1444).

---

Although, I'm wondering what exact issue are you seeing in your runtime?

If you use grpc-web npm package, it would have exposed `unaryCall` method on the client implementation, even though it's not present in the `index.d.ts` file :)

GabrieleGervasiPEL:
Sorry @sampajano It's been over a year since the question and we dropped the use of grpc web. Actually i don't remember what the problem was. Probably, reading my question, the unaryCall was not present in typings (index.d.ts) file and the generated client *ServiceClientPb.ts had a call to unaryCall, marked as red because not recognized. Maybe it's already been fixed.

sampajano:
@GabrieleGervasiPEL Hi thank you for the context! No worries at all.. And my apologies for address this late.. :)

Yeah it makes sense. :) `unaryCall` is indeed not present from `index.d.ts` but at runtime the javascript should work.. :)

[...]

PR #1444, opened 2024-06-24
[Experiment] Generate `thenableCall()` instead of `unaryCall()`

sampajano:
This came up in https://github.com/grpc/grpc-web/issues/1331

---

`thenableCall()` is what's currently exposed on the `AbstractClientBase` API.
https://github.com/grpc/grpc-web/blob/master/packages/grpc-web/index.d.ts

Which was added in the following commit for the sake of supporting "Generic API" (which no longer applies)
https://github.com/grpc/grpc-web/commit/916d6cbef42e10c301c7fd964869f3afdbcb7b17

---

`unaryCall()` is just an alias for `thenableCall()`, where it simply [calls](https://github.com/grpc/grpc-web/blob/8ab32b945ca0530222593127e9431a455eff24a8/javascript/net/grpc/web/grpcwebclientbase.js#L158-L161) the latter.

The right long-term fix is probably to get rid of `thenableCall()` in favor of `unaryCall()`, as the latter sounds like a better API (more clear).

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -689,7 +689,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Outdent();
           printer->Print(vars,
                          "}\n"
-                         "return this.client_.unaryCall(\n");
+                         "return this.client_.thenableCall(\n");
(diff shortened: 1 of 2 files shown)


sampajano:
Not planning on merging this. But will keep in mind in simplifying this going forward :)

PR #1446, opened 2024-06-25
Do not force third_party/protobuf to some old version.

benjaminp:


sampajano:
Wow thanks!!!

sampajano (approved):
Thanks so much for finding this!!

benjaminp:
This seems reasonable but I don't think it can fix the protoc_plugin `Dockerfile`. That tries to build protobuf with autotools, which has long since been removed protocolbuffers/protobuf@ed5c57a574aa24bc49432baced51aa6d01fbf38a.

sampajano:
Oh i see!!... So we might need https://github.com/grpc/grpc-web/pull/1447  afterall.. 😅

Do you know how to fix [protoc_plugin](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/docker/protoc_plugin/Dockerfile) by any chance?

It would be very appreciated if you could help too.. (I'm on a vacation soon i won't have time to fix this now..)

sampajano:
@benjaminp The APIs you were using was probably added before https://github.com/protocolbuffers/protobuf/pull/10132, right?

So maybe it would work if we re-pin it back to a version before https://github.com/protocolbuffers/protobuf/pull/10132?

sampajano:
It seems that the main reason why we need `protoc_plugin` now is it's used for releases on linux:
https://github.com/grpc/grpc-web/blob/master/.github/workflows/make-plugin-linux.yml

So if we can migrate that to Bazel as well then maybe we can kill it off entirely :)

sampajano:
@benjaminp Hi! Actually i need to revert this since we have other dependencies on that version as well.. (I should probably let CI run before merging it :))

https://github.com/grpc/grpc-web/blob/e6909cd2841061aa409e7f9365ac68b865fd8f83/packages/grpc-web/test/generated_code_test.js#L476

PR #1463, opened 2024-09-13
fix: format of typescript definition

nnnnoel:
https://www.typescriptlang.org/docs/handbook/enums.html

```proto
syntax = "proto3";

package pokerize.api.authn;

option go_package = "gen/proto/api";

import "api/common.proto";

enum RefreshTokenResponseStatus {
  refresh_token_success = 0;
  invalid_refresh_token = 1;
  refresh_token_expired = 2;
}

message UserAccount {
  string id = 1;
  string password = 2;
}
```

Before
```typescript
import * as jspb from 'google-protobuf'

import * as api_common_pb from '../api/common_pb'; // proto import: "api/common.proto"


export class UserAccount extends jspb.Message {
  getId(): string;
  setId(value: string): UserAccount;

  getPassword(): string;
  setPassword(value: string): UserAccount;

  toObject(includeInstance?: boolean): UserAccount.AsObject;
  static toObject(includeInstance: boolean, msg: UserAccount): UserAccount.AsObject;
  static serializeBinaryToWriter(message: UserAccount, writer: jspb.BinaryWriter): void;
  static deserializeBinary(bytes: Uint8Array): UserAccount;
  static deserializeBinaryFromReader(message: UserAccount, reader: jspb.BinaryReader): UserAccount;
}

export namespace UserAccount {
  export type AsObject = {
    id: string,
    password: string,
  }
}

export enum RefreshTokenResponseStatus { 
  REFRESH_TOKEN_SUCCESS = 0,
  INVALID_REFRESH_TOKEN = 1,
  REFRESH_TOKEN_EXPIRED = 2,
}
```

After
```typescript
import * as jspb from 'google-protobuf'

import * as api_common_pb from '../api/common_pb'; // proto import: "api/common.proto"


export class UserAccount extends jspb.Message {
  getId(): string;
  setId(value: string): UserAccount;

  getPassword(): string;
  setPassword(value: string): UserAccount;

  toObject(includeInstance?: boolean): UserAccount.AsObject;
  static toObject(includeInstance: boolean, msg: UserAccount): UserAccount.AsObject;
  static serializeBinaryToWriter(message: UserAccount, writer: jspb.BinaryWriter): void;
  static deserializeBinary(bytes: Uint8Array): UserAccount;
  static deserializeBinaryFromReader(message: UserAccount, reader: jspb.BinaryReader): UserAccount;
}

export namespace UserAccount {
  export type AsObject = {
    id: string;
    password: string;
  };
}

export const enum RefreshTokenResponseStatus {
  REFRESH_TOKEN_SUCCESS = 0,
  INVALID_REFRESH_TOKEN = 1,
  REFRESH_TOKEN_EXPIRED = 2,
}
```

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -778,7 +778,7 @@ void PrintProtoDtsEnum(Printer* printer, const EnumDescriptor* desc) {
   std::map<string, string> vars;
   vars["enum_name"] = desc->name();
 
-  printer->Print(vars, "export enum $enum_name$ { \n");
+  printer->Print(vars, "export const enum $enum_name$ {\n");
   printer->Indent();
   for (int i = 0; i < desc->value_count(); i++) {
     vars["value_name"] = Uppercase(desc->value(i)->name());
@@ -794,7 +794,7 @@ void PrintProtoDtsOneofCase(Printer* printer, const OneofDescriptor* desc) {
   vars["oneof_name"] = ToUpperCamel(ParseLowerUnderscore(desc->name()));
   vars["oneof_name_upper"] = Uppercase(desc->name());
 
-  printer->Print(vars, "export enum $oneof_name$Case { \n");
+  printer->Print(vars, "export const enum $oneof_name$Case {\n");
   printer->Indent();
   printer->Print(vars, "$oneof_name_upper$_NOT_SET = 0,\n");
   for (int i = 0; i < desc->field_count(); i++) {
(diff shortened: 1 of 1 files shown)


sampajano:
@nnnnoel thanks for help fixing the style!!

One request: Do you mind sharing the before v.s. after view of the generated code?

Thanks!

nnnnoel:
@sampajano I shared by PR Comment, Please see the syntax changes and check the removed space after open bracket in enum!

sampajano:
> @sampajano I shared by PR Comment, Please see the syntax changes and check the removed space after open bracket in enum!

@nnnnoel Thank you very much for providing this diff! Looks very clear! 😃

However, I've noticed that this PR not only changed format, but also changed the generated "enum" -> "const enum".

From what I read on the link you provided, there are difference in behaviors between them — _"Const enums can only use constant enum expressions and unlike regular enums they are completely removed during compilation."_

I wonder if some of our clients might be relying on the old format and can potentially be broken by this change?

nnnnoel:
Thanks for read docs @sampajano .

It has breaking change when use enum as enum key, but the typescript enum's bi-directional bind spec is not valid in grpc protocol serializer spec.
so people used it like above would not have been able to use gRPC.

sampajano:
@nnnnoel Ahh i see! Thanks for explaining! That makes sense!

But could you elaborate a bit further on what you mean by "typescript enum's bi-directional bind spec is not valid in grpc protocol serializer spec."? Could you give me a quick example maybe?

Sorry for the trouble but I want to understand a bit better before introducing a potentially breaking change 😃

nnnnoel:
@sampajano 

Before

```typescript
var key = RefreshTokenResponseStatus[RefreshTokenResponseStatus.REFRESH_TOKEN_SUCCESS]
//    ^ "REFRESH_TOKEN_SUCCESS"
```

After

```typescript
var key = RefreshTokenResponseStatus[RefreshTokenResponseStatus.REFRESH_TOKEN_SUCCESS]
//    ^ TypeError
```

sampajano:
@nnnnoel Got it!!  Thanks SO much for the clarification! Makes perfect sense now!

Thank you again for your contribution and patience! 😃

sampajano (approved):
Thanks again!! 😊

PR #1471, opened 2024-12-25
Fix webpack command in tutorials

AndiDog:
Without `./`, you get this error:

```
Module not found: Error: Can't resolve 'client.js' in '/[...]/grpc-web/net/grpc/gateway/examples/helloworld'
Did you mean './client.js'?
Requests that should resolve in the current directory need to start with './'.
```


sampajano (approved):
Thanks for helping fix this example! (Maybe webpack was updated since we've added this example, so it no longer works :)

Issue #1449, opened 2024-06-25
`protoc-plugin` image is currently broken


sampajano:
See: https://github.com/grpc/grpc-web/pull/1446#issuecomment-2187997625

The [`protoc_plugin`](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/docker/protoc_plugin/Dockerfile) image is broken since https://github.com/grpc/grpc-web/pull/1445

@benjaminp FYI :)

---

UPDATE: Temporarily disabled in https://github.com/grpc/grpc-web/pull/1451

benjaminp:
I'm afraid I can't spend a huge amount of time digging into all ways there are to build the generator at the moment in this repo. (There's also a `build.zig` file in the generator directory. What's the used for?) Possibly everything should be reverted until that's in a better place.

sampajano:
@benjaminp No worries! Appreciate your effort in trying to help anyways!!

`build.zig` is used for building for ARM architecture ([workflow](https://github.com/grpc/grpc-web/blob/master/.github/workflows/make-plugin-arm.yml). Whereas other binaries are built using platform-specific workflows: https://github.com/grpc/grpc-web/tree/master/.github/workflows

It's ok.. I think we have to move forward anyways.. I've temporarily disabled protoc_plugin in CI so BUILD is green now. But we have to fix it for a new release to be made :)

sampajano:
Note to myself: we are currently pinning the protobuf submodule version:

https://github.com/grpc/grpc-web/blob/20790fc2597971b6eaada36fbeaf5d3c95d54d93/scripts/init_submodules.sh#L19

This is for 2 reasons:

1. For protobuf JS dependencies
2. For building the generator binary

Ideally, (1) should be tackled first — by mmoving to https://github.com/protocolbuffers/protobuf-javascript, since protobuf has removed JS from its repo.

And then (2) — ensuring Makefile work as before.

sampajano:
FYI - Once we have this working we can make a test binary for https://github.com/grpc/grpc-web/issues/715

sampajano:
I think this is fixed in https://github.com/grpc/grpc-web/pull/1483, so we're unblocked for a new release :)

@Vuhag123 Thanks so much for the fix! :)

PR #1481, opened 2025-04-16
changes for build  modification and protoc-plugin

kumarakshays:
fixes-#1449


kumarakshays:
Thank you,@sampajano
Once the C++ compilation is fixed, I can create a follow-up PR to migrate the JS dependency to protobuf-javascript.

sampajano:
@Vuhag123 I've now removed proto_plugin dockerfile entirely in https://github.com/grpc/grpc-web/pull/1498. So we don't need this anymore! Thank you for working on this very much though! 😊

PR #1482, opened 2025-05-14
Update helloworld README with note on Envoy IPv4 only

LucRosset:
If the user's environment does not enable IPv6, the _envoy.yaml_ must contain the `dns_lookup_family: V4_ONLY` setting, otherwise the proxy may not work, at least for this example. See https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/run-envoy#envoy-networking for more.


sampajano:
@LucRosset Thanks a lot for adding this doc! Could you finish the EasyCLA process and we can merge it! Thanks!

sampajano (approved):

Issue #682, opened 2019-11-15
missing symbol $ in get/set functions in TypeScript output of grpc-web


renkei:
If I define a proto message with a field named "myField" then protoc generates the functions getMyfield() and setMyfield() for the related object.

What happens, if my field name is "extension"? protoc generates the functions getExtension$() and setExtension$(), it adds a $ since the functions getExtension() and setExtension() are already defined by jspb.Message for the extension api of protobuf.

grpc-web doesn't recognize this, the created TypeScript output contains getExtension() and setExtension() without the $. This causes a TypeScript compiler error, if you just change the generated d.ts files by adding the $ to the related functions then everything works as expected.

mattnathan:
I just ran across this issue, one of my messages has a field called `extension`

To reproduce:
```
syntax = "proto3";

message TestMessage {
  int32 extension = 1;
}
```

then run
`protoc -I . --js_out=import_style=commonjs,binary:. --grpc-web_out=import_style=commonjs+dts:. *.proto`

Check the `.js` and `.d.ts` and notice `proto.TestMessage.prototype.setExtension$` in the `.js` and `getExtension(): number;` in the `.d.ts`

Here's the relevant code in the js protoc compiler: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.cc#L544

renkei:
I can see that this seems to be fixed on master: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.cc#L524

Any plans to create a new release of gprc-web, e.g 1.2.2?

ericb-summit:
For some reason google removed js compilation from protoc and now it's a plugin:

https://github.com/protocolbuffers/protobuf-javascript/blob/main/generator/js_generator.cc#L525

Putting it here so when they reorg it again in 2 weeks we won't need to go down the rabbit hole

```
  if (name == "Extension" || name == "JsPbMessageId") {
    // Avoid conflicts with base-class names.
    name += "$";
  }
```

PR #1514, opened 2025-10-13
missing symbol $ in get/set functions in TypeScript output of grpc-web

Vuhag123:
[Fix682](https://github.com/grpc/grpc-web/issues/682) This PR fixes an issue where TypeScript declarations generated by `grpc-web` don't match the actual JavaScript output for reserved field names like `extension`. In JS, methods like `getExtension$()` are correctly generated to avoid conflicts, but `.d.ts` files incorrectly use `getExtension()`, causing TypeScript errors. The fix ensures consistent method naming with `$` in both JS and TS outputs.

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -485,6 +485,28 @@ string GetBasename(string filename) {
   return basename;
 }
 
+//This is for reserved method name adding $
+static bool IsReservedMethodName(const std::string& name) {
+  static const std::unordered_set<std::string> reserved = {
+    "extension", 
+    "Extension" 
+  };
+
+  std::string lower_name = name;
+  std::transform(lower_name.begin(), lower_name.end(), lower_name.begin(), ::tolower);
+
+  bool is_reserved = reserved.count(lower_name) > 0;
+  return is_reserved;
+}
+
+static std::string SafeAccessorName(const std::string& name) {
+  std::string result = name;
+  if (IsReservedMethodName(name)) {
+    result += "$";
+  }
+  return result;
+}
+
 // Finds all message types used in all services in the file. Return results as a
 // map of full names to descriptors to get sorted results and deterministic
 // build outputs.
@@ -823,7 +845,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Indent();
   for (int i = 0; i < desc->field_count(); i++) {
     const FieldDescriptor* field = desc->field(i);
-    vars["js_field_name"] = JSFieldName(field);
+
+    string base_js_field_name = JSFieldName(field);
+    string accessor_name = SafeAccessorName(base_js_field_name);
+
+    vars["js_field_name"] = accessor_name;
     vars["js_field_type"] = JSFieldType(field, file);
     if (field->type() != FieldDescriptor::TYPE_MESSAGE ||
         field->is_repeated()) {
@@ -855,7 +881,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
       printer->Print(vars, "clear$js_field_name$(): $class_name$;\n");
     }
     if (field->is_repeated() && !field->is_map()) {
-      vars["js_field_name"] = JSElementName(field);
+
+      string base_elem_name = JSElementName(field);
+      string accessor_elem_name = SafeAccessorName(base_elem_name);
+
+      vars["js_field_name"] = accessor_elem_name;
       vars["js_field_type"] = JSElementType(field, file);
       if (field->type() != FieldDescriptor::TYPE_MESSAGE) {
         printer->Print(vars,
@@ -900,11 +930,14 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Indent();
   for (int i = 0; i < desc->field_count(); i++) {
     const FieldDescriptor* field = desc->field(i);
-    string js_field_name = CamelCaseJSFieldName(field);
-    if (IsReserved(js_field_name)) {
-      js_field_name = "pb_" + js_field_name;
+    
+    string base_js_field_name = CamelCaseJSFieldName(field);
+    string accessor_name = SafeAccessorName(base_js_field_name);
+
+    if (IsReserved(accessor_name)) {
+      accessor_name = "pb_" + accessor_name;
     }
-    vars["js_field_name"] = js_field_name;
+    vars["js_field_name"] = accessor_name;
     vars["js_field_type"] = AsObjectFieldType(field, file);
     if (!field->has_presence()) {
       printer->Print(vars, "$js_field_name$: $js_field_type$;\n");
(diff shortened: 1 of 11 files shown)


sampajano (commented):
@Vuhag123 Thanks for the change!!

A few requests:

1. Could you add a bit more context in the bug description? i.e. what issue is it trying to fix? Why does it matter? 
2. It would be preferable to have some testing for this, if not, could you specify how do you personally test it and what does the Before v.s. After gencode look like? 
3. (minor) Could you clean up the git commit history? (it has a bunch of release related commit it seems, which doesn't seem necessary to be included :)) 

Thanks!

Vuhag123:
> @Vuhag123 Thanks for the change!!
> 
> A few requests:
> 
> 1. Could you add a bit more context in the bug description? i.e. what issue is it trying to fix? Why does it matter?
> 2. It would be preferable to have some testing for this, if not, could you specify how do you personally test it and what does the Before v.s. After gencode look like?
> 3. (minor) Could you clean up the git commit history? (it has a bunch of release related commit it seems, which doesn't seem necessary to be included :))
> 
> Thanks!

1. Could you add a bit more context in the bug description? i.e. what issue is it trying to fix? Why does it matter? -- **Added description.**
2. It would be preferable to have some testing for this, if not, could you specify how do you personally test it and what does the Before v.s. After gencode look like? -- **I did a manual test, I generated TypeScript code using the protoc command with the modified generator. I verified that the methods for fields like extension correctly included the $ suffix (i.e., getExtension$() and setExtension$()) in the generated .d.ts file. The test ensures there are no conflicts with jspb.Message methods and that the TypeScript code compiles without errors.  I will attach before and after .ts files here.** 
3. (minor) Could you clean up the git commit history? (it has a bunch of release related commit it seems, which doesn't seem necessary to be included :)) -- **will remove it.**

**Before:**
------------
Before

**After**
------------------
After

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -823,7 +844,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Indent();
   for (int i = 0; i < desc->field_count(); i++) {
     const FieldDescriptor* field = desc->field(i);
-    vars["js_field_name"] = JSFieldName(field);
+
+    string base_js_field_name = JSFieldName(field);
+    string accessor_name = SafeAccessorName(base_js_field_name);
+
+    vars["js_field_name"] = accessor_name;
```
sampajano: nit: maybe just simplify like this? :)

vars["js_field_name"] = SafeAccessorName(JSFieldName(field));

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -855,7 +880,11 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
       printer->Print(vars, "clear$js_field_name$(): $class_name$;\n");
     }
     if (field->is_repeated() && !field->is_map()) {
-      vars["js_field_name"] = JSElementName(field);
+
+      string base_elem_name = JSElementName(field);
+      string accessor_elem_name = SafeAccessorName(base_elem_name);
+
+      vars["js_field_name"] = accessor_elem_name;
```
sampajano: same as above.

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -900,11 +929,14 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
...
   for (int i = 0; i < desc->field_count(); i++) {
     const FieldDescriptor* field = desc->field(i);
-    string js_field_name = CamelCaseJSFieldName(field);
-    if (IsReserved(js_field_name)) {
-      js_field_name = "pb_" + js_field_name;
+    
+    string base_js_field_name = CamelCaseJSFieldName(field);
+    string accessor_name = SafeAccessorName(base_js_field_name);
+
+    if (IsReserved(accessor_name)) {
```
sampajano: Just to double check: Is it still necessary to add "$" when we already add "pb_"?

Does the underlying JS code add both?

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -485,6 +485,27 @@ string GetBasename(string filename) {
...
+static bool IsReservedMethodName(const std::string& name) {
+  static const std::unordered_set<std::string> reserved = {
+    "extension",
+    "JsPbMessageId"
+  };
+
+  std::string lower_name = name;
+  std::transform(lower_name.begin(), lower_name.end(), lower_name.begin(), ::tolower);
+
+  return reserved.count(lower_name) > 0;
```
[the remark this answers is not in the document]
sampajano: This makes sense. @Vuhag123 Can you fix?

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -900,11 +929,14 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
...
-      js_field_name = "pb_" + js_field_name;
+    
+    string base_js_field_name = CamelCaseJSFieldName(field);
+    string accessor_name = SafeAccessorName(base_js_field_name);
+
+    if (IsReserved(accessor_name)) {
+      accessor_name = "pb_" + accessor_name;
     }
-    vars["js_field_name"] = js_field_name;
+    vars["js_field_name"] = accessor_name;
```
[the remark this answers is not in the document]
sampajano: @Vuhag123 this seems to make sense too. can you double check? (I had a similar comment.)

sampajano (commented):
Thanks @Vuhag123! Looks like a good fix!

But there's a few comments I've added. Also some Github Copilot comments seems to make sense too (2 looks like potential bugs). Can you go over them and evaluate and respond? Thanks!! 😃

javascript/net/grpc/web/generator/grpc_generator.cc
Vuhag123: @sampajano ,The $ suffix in SafeAccessorName() is still needed. "pb_" handles JS property name safety, while the $ ensures TypeScript field names in .d.ts are valid identifiers. Removing $ would reintroduce the TS compilation issues, so both are required.

Vuhag123: yes checked and removed it and thanks for suggesting it.

sampajano (commented):
Thanks Vuha!!

Another seeming bug is here, could you take a look too? Thanks!

https://github.com/grpc/grpc-web/pull/1514/files#r2430747646

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -900,10 +922,12 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Indent();
   for (int i = 0; i < desc->field_count(); i++) {
     const FieldDescriptor* field = desc->field(i);
+
```
sampajano, line 925: nit: might revert these empty line changes (since no longer changing logics around it).

sampajano (approved):
Great! Thanks for the change, @Vuhag123 ! :)

Issue #1097, opened 2021-06-17
Typescript type definition for proto constructor does not allow arguments


forrestli74:
In short, for constructor of messages, why does the *_pb.js version take an argument and the *_pb.d.ts does not take any argument?

I compiled my proto file to typescript by running something like this:
protoc \
	--js_out=import_style=commonjs:. \
	--grpc-web_out=import_style=commonjs+dts,mode=grpcweb:. FILE

In *_pb.js, the constructor of message looks something like this:
```
function(opt_data) {
  jspb.Message.initialize(this, opt_data, 0, -1, null, null);
};
```
But in *_pb.d.ts, the constructor of the message only have default constructor, which takes no arguments.

I would expect in *_pb.d.ts, it has a constructor like this:
```
export class Example extends jspb.Message {
  constructor(opt_data: Example.AsObject);
}
```

The reason I want it is because I want to construct a message from a generic native object.

Vuhag123:
Fixed the issue and PR is merged.

PR #1515, opened 2025-10-24
Fix 1097 - typescript: add missing constructor definition in generated message classes.

Vuhag123:
Fix (https://github.com/grpc/grpc-web/issues/1097) - typescript: add missing constructor definition in generated message classes. 

After_1097

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -485,6 +485,26 @@ string GetBasename(string filename) {
   return basename;
 }
 
+//Adds $ suffix to reserved method names to avoid conflicts.
+static bool IsReservedMethodName(const std::string& name) {
+  static const std::unordered_set<std::string> reserved = {
+    "extension",
+    "jspbmessageid"
+  };
+
+  std::string lower_name = Lowercase(name);
+
+  return reserved.count(lower_name) > 0;
+}
+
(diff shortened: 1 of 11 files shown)


javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -821,6 +821,9 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
 
   printer->Print(vars, "export class $class_name$ extends jspb.Message {\n");
   printer->Indent();
+
+  printer->Print(vars, "constructor(opt_data?: $class_name$.AsObject);\n");
```
sampajano, line 845: @Vuhag123 Thanks! for the change!

One request: Can you provide an example output in the PR description, and try verifying this change works (via a typescript example)?

Vuhag123: I have verified with typescript example it is working, i am attaching the images here too
Below is existing one:
Before_1097
After fix, 
After_1097

sampajano: Oh nice! This is great! Thanks for showing the change in action! This would be helpful indeed! Thanks Vuha :)

sampajano (approved):
Thanks so much for the change, @Vuhag123 ! :)

asheshvidyut (approved):
LGTM

Issue #944, opened 2020-08-31
Generated Typescript code is not compiling with type checking enabled 


uwemaurer:
The generated typescript code for the GRPC client does not compile when the type checking is enabled.
This change [Disable static checkers on generated js files](https://github.com/grpc/grpc-web/commit/87209cb94bb5047509ce3f993b6d2ec78da7714e) added `//@ts-nocheck` which allows the typescript compiler starting from version 3.7 to accept the code, but with older versions (which ignore this comment option) it fails.

There are two problems in the generated code: The code calls `unaryCall` but this method got removed from the interface `AbstractClientBase` already.
Also it passes `MethodInfo` where `MethodDescriptor` is expected.

Is there a reason why it is like this? Otherwise I suggest to remove this  `//@ts-nocheck`  again and let the generator produce code which passes the type checking.

As an alternative the `index.d.ts` could be updated to correctly declare the methods as they are currently used.

```
   export class AbstractClientBase {
+    /** @deprecated use thenableCall */
+    unaryCall<REQ, RESP> (
+      method: string,
+      request: REQ,
+      metadata: Metadata,
+      methodDescriptor: MethodDescriptor<REQ, RESP> | MethodInfo<REQ, RESP>
+    ): Promise<RESP>;
+
     thenableCall<REQ, RESP> (
       method: string,
       request: REQ,
       metadata: Metadata,
-      methodDescriptor: MethodDescriptor<REQ, RESP>
+      methodDescriptor: MethodDescriptor<REQ, RESP> | MethodInfo<REQ, RESP>
     ): Promise<RESP>;
 
     rpcCall<REQ, RESP> (
       method: string,
       request: REQ,
       metadata: Metadata,
-      methodDescriptor: MethodDescriptor<REQ, RESP>,
+      methodDescriptor: MethodDescriptor<REQ, RESP> | MethodInfo<REQ, RESP>,
       callback: (err: Error, response: RESP) => void
     ): ClientReadableStream<RESP>;
 
@@ -30,7 +38,7 @@
       method: string,
       request: REQ,
       metadata: Metadata,
-      methodDescriptor: MethodDescriptor<REQ, RESP>
+      methodDescriptor: MethodDescriptor<REQ, RESP>  | MethodInfo<REQ, RESP>
     ): ClientReadableStream<RESP>;
   }
```

vladfr:
Also, am I the only one who's having trouble with importing the `REQ` and `RESP` types here?

[...]

PR #1521, opened 2026-01-19
fix-944 - Added unary Call in abstractclientbase.

Vuhag123:
fix-[944](https://github.com/grpc/grpc-web/issues/944) Updates the generated gRPC-Web client to add a typed unaryCall implementation in AbstractClientBase, improving support for Promise-style unary calls. This makes the client more ergonomic for TypeScript usage without changing any existing runtime behavior. It also adjusts the generated constructor types to improve compatibility with generic instantiation patterns.

### DIFF
--- a/javascript/net/grpc/web/generator/grpc_generator.cc
+++ b/javascript/net/grpc/web/generator/grpc_generator.cc
@@ -710,7 +710,7 @@ void PrintTypescriptFile(Printer* printer, const FileDescriptor* file,
           printer->Outdent();
           printer->Print(vars,
                          "}\n"
-                         "return this.client_.thenableCall(\n");
+                         "return this.client_.unaryCall(\n");
           printer->Print(vars,
                          "this.hostname_ +\n"
                          "  '/$package_dot$$service_name$/$method_name$',\n"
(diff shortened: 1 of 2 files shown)


sampajano:
@Vuhag123 Thanks for the change!!

Although, I'm wondering if there's a way to make `unaryCall` to work with type checking?

For context: `thenableCall` was a later addition used to facilitate a migration which later turned out to be unnecessary.. :)

So if we were to keep one between `thenableCall` and `unaryCall`, we should keep the latter — it will help us reduce maintenance cost in the long run.

javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -842,8 +842,8 @@ void PrintProtoDtsMessage(Printer* printer, const Descriptor* desc,
   printer->Print(vars, "export class $class_name$ extends jspb.Message {\n");
   printer->Indent();
 
-  printer->Print(vars, "constructor(opt_data?: $class_name$.AsObject);\n");
-  
+  printer->Print(vars, "constructor();\n" "constructor(opt_data?: $class_name$.AsObject);\n");
```
sampajano, line 845: Curious why is this necessary if the original constructor already is taking an optional argument?

sampajano (approved):
Thanks Vuha! Adding `unaryCall` looks good to me!

Please update PR name and descriptions too 😃

javascript/net/grpc/web/generator/grpc_generator.cc
Vuhag123: Even though the param is optional, TS still treats the constructor as “takes a specific arg”. Some generic code expects to be able to do new T() with no args, so adding the no-arg overload just makes the type more flexible. No runtime change, just smoother TS compatibility

sampajano: Ah i see — so it's for TS usability.

Although for that purpose alone, i believe it would work whether we make this JS change or not.

If you add the default constructor in `.d.ts` file, then it should work with this generated JS code even without this change.

(What i mean is, TS code should not need to read into actual JS implementation — and the default constructor should work with the original code)

Could you try it maybe? Thanks!

PR #1534, opened 2026-06-09
TypeScript Migration: Core Library and Build Setup

Vuhag123:
- Migrates core runtime library files to TypeScript.
  - Updates search-path and dependency overrides to use packages/closure-wrapper.
  - Exposes MockXhrIo definitions and the MethodInfo alias in the closure-net dependency.
  - Coordinates Dockerfile.ts and docker-compose.yml for safe test runs.
  - 
TESTED=Ran 'docker compose build ts-prereqs' and verified that all tsc-tests and browser echo tests pass successfully


net/grpc/gateway/examples/echo/ts-example/client.ts
```diff
@@ -17,14 +17,14 @@
  */
 
 import * as grpcWeb from 'grpc-web';
-import * as $ from 'jquery';
+import $ from 'jquery';
 
 // Uncomment either one of the following:
```
mbissa, line 22: why are we using both? would it not make sense to remove this and keep only typescript style?

packages/closure-wrapper/dist/grpc_web_blob_es2022.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
```
sampajano: In general, it is preferable not to commit compiled outputs (like `dist/`) to version control to avoid unnecessary diff noise. Can these be built instead on the fly using a pre-install/prepare script as part of the build pipeline?

packages/grpc-web/package.json
```diff
@@ -24,8 +24,14 @@
     "test-jsunit": "./scripts/generate_test_files.sh && ./scripts/run_jsunit_tests.sh && rm -rf ./generated"
   },
   "license": "Apache-2.0",
+  "dependencies": {
+    "@grpc-web/closure-wrapper": "file:../closure-wrapper"
```
sampajano: Thanks for working on this, @Vuhag123! But I believe we should avoid adding `@grpc-web/closure-wrapper` as a runtime `dependency` for `grpc-web` users. Historically, `grpc-web` has maintained a minimal runtime dependency footprint (only depending on `google-protobuf`), which is a key reason it is so widely adopted and easy to integrate.

We should not introduce a new runtime dependency unless absolutely necessary, which is not the case here. Since `@grpc-web/closure-wrapper` is an internal utility package local to this monorepo and only used by `grpc-web`, it is a perfect candidate to be a `devDependency` that is inlined/bundled directly into the compiled distribution bundles of `grpc-web` (both CJS and ESM).

As side evidence of the fragility of keeping it as a runtime dependency, the current commit in `packages/closure-wrapper` actually misses the compiled ESM folder (`dist/esm/`), which would immediately break the build with a "Module not found" error for any modern ESM-based application consuming the package. Inlining it completely bypasses these types of dependency resolution issues.

Could you please move `@grpc-web/closure-wrapper` to `devDependencies` and configure Rollup to bundle/inline it during the build of `grpc-web`?

sampajano (commented):
@Vuhag123 Thank you SO MUCH for working on this!! Glad it's all working!! 😃😃

Although, I have a major concern regarding how closure-wrapper is taken in as an actual "dependency" and I left a comment there.

Could you take a look? 

Thanks!! 😊

packages/closure-wrapper/dist/grpc_web_blob_es2022.js
Vuhag123: Addressed this issue can check recent commit

packages/grpc-web/package.json
Vuhag123: Thanks for the great feedback, @sampajano! I completely agree with keeping the runtime dependency footprint minimal, so I've moved @grpc-web/closure-wrapper to devDependencies in packages/grpc-web/package.json. I also configured Rollup (rollup.config.js) to automatically bundle and inline closure-wrapper directly into both CommonJS (index.js) and ESM (index.esm.js) distribution bundles during the build. Finally, I updated closure-wrapper to generate both CJS and ESM outputs on install via a prepare script and removed the precompiled dist/ files from git. Everything builds cleanly and all Docker tests are passing green!

packages/closure-wrapper/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
sampajano, line 1: Mixing different package managers (Yarn and NPM) within the same monorepo is generally avoided as it is considered an anti-pattern. Having split lockfiles (`yarn.lock` and `package-lock.json`) can easily lead to inconsistent dependency resolution across packages, complicates CI cache configurations, and adds toolchain friction for contributors.

I wonder if it's definitely necessary to use Yarn?

To keep the repository uniform and follow best practices, would it make sense to standardize `@grpc-web/closure-wrapper` on NPM by removing `yarn.lock` and using NPM for dependency management and builds instead?

packages/closure-wrapper/blob/package.json
```diff
@@ -0,0 +1,7 @@
+  {
```
sampajano, line 1: In `packages/closure-wrapper/blob/package.json`, `typings` references a deep relative path:
```json
"typings": "../../../node_modules/closure-net/grpc_web/grpc_web_blob_types.d.ts"
```
This assumes the dependency will always be hoisted to the root `node_modules`, which might be fragile if the package is installed in environments with strict isolation (like pnpm or Yarn PnP).

Since the build step copies `grpc_web_blob_types.d.ts` into the package's local `dist/` directory anyway, would it be a bit safer and more self-contained to reference the local copy instead (e.g., `"typings": "../dist/grpc_web_blob_types.d.ts"`)?

packages/grpc-web/package.json
```diff
@@ -24,8 +24,14 @@
     "test-jsunit": "./scripts/generate_test_files.sh && ./scripts/run_jsunit_tests.sh && rm -rf ./generated"
   },
   "license": "Apache-2.0",
+  "dependencies": {
```
sampajano: It looks like the newly added TypeScript unit tests (`src/typescript/grpcwebclientbase_test.ts` and `src/typescript/grpcwebstreamparser_test.ts`) are not currently executed by the build scripts or the Docker-based test suite. The `"test-mocha"` script in `packages/grpc-web/package.json` is configured to target `"./test/**/*_test.js"`, which misses these new `.ts` tests located under the root `src/` directory.

What do you think about adding a test script (or updating the existing one) to compile and run these new unit tests, so we can ensure they are part of the CI pipeline?

net/grpc/gateway/docker/ts_prereqs/Dockerfile
```diff
@@ -0,0 +1,26 @@
+  # 1. Start from the image built by the original prereqs file!
```
sampajano, line 1: I noticed that the Dockerfile for `ts-prereqs` is named `dockerfile.ts` and placed inside the `src/typescript/` source folder. This seems to cause IDEs and the compiler to attempt to parse it as TypeScript, and placing configuration files under `src/` deviates slightly from the project's layout conventions (where other Dockerfiles live in a dedicated `docker/` folder).

What do you think about renaming this to `Dockerfile` and moving it to a directory like `net/grpc/gateway/docker/ts_prereqs/Dockerfile`?

packages/grpc-web/index.d.ts
```diff
@@ -3,6 +3,9 @@ declare module "grpc-web" {
   export interface Metadata { [s: string]: string; }
 
   export class AbstractClientBase {
+
+    static MethodInfo: new (...args: any[]) => any;
```
sampajano: Thanks for  adding back legacy exports like `Error` (alias for `RpcError`) and `MethodInfo` (alias for `MethodDescriptor`) to maintain backward compatibility!

I was wondering, are we planning to support these compatibility aliases indefinitely, or should we mark them as `@deprecated` to suggest migrating to the newer names in a future release?

sampajano (commented):
@Vuhag123 Hi Vuha! Thanks so much for addressing the comments! Looking GREAT now! Just a few smaller comments! Thanks!! 😊

packages/grpc-web/package.json
sampajano: Thanks so much for fixing this, @Vuhag123 ! Looking Great! Appreciate it! 😊

packages/closure-wrapper/package.json
Vuhag123: I completely agree that mixing package managers is usually an anti-pattern, but Yarn is strictly necessary here because the upstream `closure-net` repository lacks a root `package.json`. Modern NPM (v9+) fails with an `ENOENT` error when installing sourceless Git bundles, whereas Yarn fetches them flawlessly. We are following the exact architectural precedent set by Google's official wrapper at https://github.com/google/closure-net/tree/main/firebase, which also relies on Yarn for this exact reason. To ensure CI stability and prevent cache collisions during Docker builds, I added `clean: true` and `copyOnce: true` to our Rollup config.

packages/closure-wrapper/blob/package.json
Vuhag123: Great catch! Relying on root `node_modules` hoisting is definitely fragile in strict setups like PNPM or Yarn PnP. I updated `blob/package.json` to point directly to our local copy (`"typings": "../dist/grpc_web_blob_types.d.ts"`). I also updated `main` in `closure-wrapper` to point to `dist/` so both TypeScript definitions and CJS/ESM imports resolve perfectly out of the box!

net/grpc/gateway/docker/ts_prereqs/Dockerfile
Vuhag123: Totally agree! Having a Dockerfile ending in `.ts` inside the source tree was definitely confusing IDEs and `tsc`. E.g., I moved it exactly where you suggested to `net/grpc/gateway/docker/ts_prereqs/Dockerfile`, renamed it to a clean `Dockerfile`, and updated `docker-compose.yml` to match. The TypeScript source tree is now completely clean of configs!

packages/grpc-web/index.d.ts
Vuhag123: That makes total sense! I added clean JSDoc `/** @deprecated */` tags to both `MethodInfo` and `Error` in `packages/grpc-web/index.d.ts`. This maintains perfect backward compatibility for now while gently encouraging users and IDEs to migrate to the new names (`MethodDescriptor` and `RpcError`) in future releases!

net/grpc/gateway/docker/ts_prereqs/Dockerfile
sampajano: Gerat Thanks!

packages/closure-wrapper/blob/package.json
sampajano: Great, thanks!

packages/closure-wrapper/package.json
sampajano: Ah ok! thanks for explaning!

btw, it seems that there are a good amount of design decisions you made in this PR. Do you mind mentioning the most important design decisions you took in the PR description? thanks!

packages/grpc-web/package.json
sampajano: @Vuhag123 Seems this is done now? Thanks! :)

packages/grpc-web/index.d.ts
sampajano: Great. Thanks! :)

sampajano (approved):
@Vuhag123 Thanks so much for iterating on the PR, Vuha! Looks great to me now! Thanks so much for working on this again! Appreciate it very much! 😊

mbissa (approved):

net/grpc/gateway/docker/prereqs/Dockerfile
```diff
@@ -87,9 +87,19 @@ RUN bazel build javascript/net/grpc/web/generator:protoc-gen-grpc-web && \
...
+    protoc -I=./src/typescript/proto -I=/usr/local/include \
+    ./src/typescript/proto/status.proto \
+    --js_out=import_style=commonjs:./src/typescript/generated \
+    --grpc-web_out=import_style=commonjs+dts,mode=grpcwebtext:./src/typescript/generated
+
+RUN cd ./packages/closure-wrapper && \
+  yarn install && \
+  yarn build && \
+  cd ../grpc-web && \
+  npm install --ignore-scripts && \
```
mbissa, line 102: add a comment to explain --ignore-scripts

PR #1536, opened 2026-07-23
  fix: pin typescript (~5.5.4) and @types/node (^20.14.0) for deterministic Docker CI builds      

Vuhag123:
- **Fix Node 20 ESM Crash:** Pinned `"typescript": "~5.5.4"` across package files to prevent `ERR_UNKNOWN_FILE_EXTENSION` when running `npx tsc` inside Docker on fresh CI VMs.
  - **Fix Node Type Errors:** Pinned `"@types/node": "^20.14.0"` to avoid missing `IteratorObject` compiler errors from floating `"latest"` definitions.
  - **Configure tsconfig:** Enabled `esModuleInterop` and disabled `strictNullChecks` in `ts-example/tsconfig.json` for CommonJS test compatibility.

### DIFF
--- a/net/grpc/gateway/examples/echo/ts-example/tsconfig.json
+++ b/net/grpc/gateway/examples/echo/ts-example/tsconfig.json
@@ -5,7 +5,12 @@
     "strict": true,
     "allowJs": true,
     "outDir": "./dist",
-    "types": ["node"]
+    "types": [
+      "node"
+    ],
+    "esModuleInterop": true,
+    "noImplicitAny": false,
+    "strictNullChecks": false
   },
   "include": [
     "client.ts",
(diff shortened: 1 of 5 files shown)


net/grpc/gateway/examples/echo/ts-example/tsconfig.json
```diff
@@ -5,7 +5,12 @@
     "strict": true,
     "allowJs": true,
     "outDir": "./dist",
-    "types": ["node"]
+    "types": [
+      "node"
+    ],
+    "esModuleInterop": true,
+    "noImplicitAny": false,
+    "strictNullChecks": false
```
mbissa, line 13: add a github ticket to turn "strictNullChecks": true later when we can. This is ok for now since we are migrating, but we should ideally have strict checking here.

mbissa (approved):

net/grpc/gateway/examples/echo/ts-example/tsconfig.json
Vuhag123: ok sure will change later with new ticket

PR #1540, opened 2026-07-28
chore: bump version to 2.1.0

Vuhag123:


javascript/net/grpc/web/generator/grpc_generator.cc
```diff
@@ -81,7 +81,7 @@ const char* kKeyword[] = {
 };
 
 // Edit the version here prior to release
-static const std::string GRPC_WEB_VERSION = "2.0.2";
```
mbissa, line 84: Why are we moving to 3 and not 2.1?

Vuhag123: We are moving to **`3.0.0`** because this release marks the complete migration of the core library from legacy JavaScript and Google Closure Library to modern **TypeScript**.

mbissa (approved):

mbissa (changes requested):
I think we should do a minor version upgrade, since the changes are not breaking changes.

Vuhag123:
@mbissa  Agreed! Since there are no breaking changes, I just updated the version from 3.0.0 to 2.1.0. Thanks!

mbissa (approved):

PR #1547, opened 2026-08-20
limit maxReceiveMessageSize to 4MB to have parity with other runtimes

mbissa:
Release Notes: TBD

### DIFF
--- a/javascript/net/grpc/web/grpcwebclientreadablestream.js
+++ b/javascript/net/grpc/web/grpcwebclientreadablestream.js
@@ -65,8 +65,10 @@ class GrpcWebClientReadableStream {
   /**
    * @param {!GenericTransportInterface} genericTransportInterface The
    *   GenericTransportInterface
+   * @param {number=} maxMessageLength The maximum allowed inbound message
+   *   length in bytes, forwarded to the stream parser. Defaults to 4 MB.
    */
-  constructor(genericTransportInterface) {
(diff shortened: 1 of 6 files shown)


javascript/net/grpc/web/grpcwebstreamparser.js
```diff
@@ -212,9 +224,22 @@ class GrpcWebStreamParser {
...
+      // Multiply rather than left-shift: the JS `<<` operator coerces to a
+      // signed 32-bit int, which would misread any length >= 2^31 as negative.
+      parser.length_ = (parser.length_ * 256) + b;
 
       if (parser.countLengthBytes_ == 4) {  // no more length byte
+        // Reject an oversized (possibly attacker-controlled) length before
+        // allocating, so a bogus frame header cannot force a huge allocation.
+        if (parser.maxMessageLength_ > 0 &&
+            parser.length_ > parser.maxMessageLength_) {
+          throw new Error(
```
asheshvidyut: This files has `parser.error_` used. Should we use the same, rather than `throw new Error` ?

javascript/net/grpc/web/grpcwebclientreadablestream.js
asheshvidyut: Should we use `StatusCode.RESOURCE_EXHAUSTED` when we see the limit reached?

We need to check the status code for Go/Java/Core as well, and make the similar change here.

Vuhag123 (approved):
approved.

